PR #14, opened 2019-10-10
Use the cocina DRO types

jcoyne:
Ref  https://github.com/sul-dlss/dor-services-client/issues/106

### DIFF
--- a/lib/cocina/models/dro.rb
+++ b/lib/cocina/models/dro.rb
@@ -6,6 +6,23 @@ module Cocina
   module Models
     # A digital repository object.  See https://github.com/sul-dlss-labs/taco/blob/master/maps/DRO.json
     class DRO < Dry::Struct
+      TYPES = %w[
+        http://sdr.sul.stanford.edu/models/sdr3-object.jsonld
+        http://sdr.sul.stanford.edu/models/sdr3-3d.jsonld
+        http://sdr.sul.stanford.edu/models/sdr3-agreement.jsonl
+        http://sdr.sul.stanford.edu/models/sdr3-book.jsonld
+        http://sdr.sul.stanford.edu/models/sdr3-document.jsonld
+        http://sdr.sul.stanford.edu/models/sdr3-geo.jsonld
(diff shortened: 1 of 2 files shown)


lib/cocina/models/dro.rb
```diff
@@ -6,6 +6,23 @@ module Cocina
   module Models
     # A digital repository object.  See https://github.com/sul-dlss-labs/taco/blob/master/maps/DRO.json
     class DRO < Dry::Struct
+      DRO_TYPES = %w[
```
mjgiarlo: Would you consider naming this TYPES since it's already in the DRO context?

jcoyne: yep, changed

lib/cocina/models/dro.rb
```diff
@@ -6,6 +6,23 @@ module Cocina
...
+        http://sdr.sul.stanford.edu/models/sdr3-geo.jsonld
+        http://sdr.sul.stanford.edu/models/sdr3-image.jsonld
+        http://sdr.sul.stanford.edu/models/sdr3-page.jsonld
+        http://sdr.sul.stanford.edu/models/sdr3-photograph.jsonld
+        http://sdr.sul.stanford.edu/models/sdr3-manuscript.jsonld
+        http://sdr.sul.stanford.edu/models/sdr3-map.jsonld
+        http://sdr.sul.stanford.edu/models/sdr3-media.jsonld
+        http://sdr.sul.stanford.edu/models/sdr3-track.jsonld
+        http://sdr.sul.stanford.edu/models/sdr3-webarchive-binary.jsonld
+        http://sdr.sul.stanford.edu/models/sdr3-webarchive-seed.jsonld
```
mjgiarlo: I'm not super stoked about keeping the `sdr3` name around, but I realize that making that change *here* is not the right place. (And I don't know how much support we have, nor what process we'd use, for tweaking the modeling documentation.) Just registering that with you to see what you think. Approving PR.

jcoyne: I'd be okay with me if we drop the `3`

PR #15, opened 2019-10-11
Add a factory method for building dro/collection

jcoyne:


lib/cocina/models.rb
```diff
@@ -25,8 +25,23 @@ def camelize(basename, _abspath)
...
+    # @raises [RuntimeError] if a valid type is not found in the data
+    # @raises [KeyError] if a type field cannot be found in the data
+    def self.build(dyn)
+      case dyn.fetch('type')
+      when *DRO::TYPES
+        DRO.from_dynamic(dyn)
+      when *Collection::TYPES
+        Collection.from_dynamic(dyn)
+      else
+        raise "Unknown type: '#{dyn.fetch('type')}'"
```
mjgiarlo: Worth raising the custom error type defined on line 30 here so clients get a more specific error? Or creating a new error subtype for unknown types?

jcoyne: Yes, fixed.

PR #27, opened 2020-01-14
Point at TravisCI.com for badge and service

mjgiarlo:
## Why was this change made?

To have a single build instead of multiple builds. I also removed the travis-ci.org webhook and installed the codeclimate coverage webhook.

yes, the badge

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -17,3 +17,7 @@ after_script:
 
 notifications:
   email: false
+
+env:
+  global:
+    - CC_TEST_REPORTER_ID=882e943342bed5285a6078be9a41b7738ae7ebd17458c08414581cbece99b8e6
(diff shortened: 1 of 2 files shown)


ndushay:
thanks for this, Mike.  Do we have much at travis.com?  How was I to know this?

mjgiarlo:
@ndushay I only knew this because I saw Travis running two sets of builds. After more poking around than I care to admit... I finally figured out that `master` was building on travis-ci.com and other branches were building on travis-ci.org. I believe the .org integration was set up via the (legacy?) webhook method, and the .com integration is done through a GitHub "App" which is tied to our @sul-dlss org. Looks like a bunch of stuff is building on the .com infrastructure: https://travis-ci.com/sul-dlss I don't know what the history is here.

ndushay:
oy.  thanks.

PR #33, opened 2020-01-29
Allow DRO to contain filesets

jcoyne:
## Why was this change made?
This matches what we're doing in sdr-api and is derivable from contentMetadata.xml

yes

### DIFF
--- a/lib/cocina/models/file_set.rb
+++ b/lib/cocina/models/file_set.rb
@@ -17,11 +17,11 @@ module Cocina
 
       # Structural sub-schema for the FileSet
       class Structural < Dry::Struct
-        attribute :contains, Types::Strict::Array.of(Types::Coercible::String).meta(omittable: true)
+        attribute :contains, Types::Strict::Array.of(Cocina::Models::File).meta(omittable: true)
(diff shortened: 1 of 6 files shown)


lib/cocina/models/file_set.rb
```diff
@@ -17,11 +17,11 @@ class Identification < Dry::Struct
 
       # Structural sub-schema for the FileSet
       class Structural < Dry::Struct
-        attribute :contains, Types::Strict::Array.of(Types::Coercible::String).meta(omittable: true)
+        attribute :contains, Types::Strict::Array.of(Cocina::Models::File).meta(omittable: true)
```
mjgiarlo: Can `Cocina::Models::` be dropped from lines 20 and 24? The explicit namespace does not appear to be required in the changing to dro.rb above.

jcoyne: I just don't want anyone to think this is `::File`, so I'm being explicit here.

mjgiarlo: Aha. Makes sense.

PR #35, opened 2020-01-30
Collections should have release tags

jcoyne:
## Why was this change made?

So that we can communicate about release tags on collections. This will allow us to remove Fedora calls in the release workflow.

n/a

### DIFF
--- a/docs/maps/Collection.json
+++ b/docs/maps/Collection.json
@@ -125,6 +125,13 @@
           "description": "Administrative or Internal project this resource is a part of.",
           "type": "string"
         },
+        "releaseTags": {
+          "description": "Tags for release",
+          "type": "array",
+          "items": {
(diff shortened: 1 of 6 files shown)


docs/maps/Collection.json
```diff
@@ -125,6 +125,13 @@
           "description": "Administrative or Internal project this resource is a part of.",
           "type": "string"
         },
+        "releaseTags": {
+          "description": "Tags for release",
```
justinlittman: A tad more description would be helpful. I have no idea what a release tag is.

jcoyne: @justinlittman that makes two of us.  This currently is an XML node that gets made in identityMetadata when you press the "Manage Release" button in Argo.  I've asked @andrewjbtw to help explain it more.

PR #56, opened 2020-02-13
Add the access property for items and collections

jcoyne:
## Why was this change made?

This is required for registering from argo as the form lets you provide access.

n/a

### DIFF
--- a/lib/cocina/models/collection.rb
+++ b/lib/cocina/models/collection.rb
@@ -16,6 +16,8 @@ module Cocina
 
       # Subschema for access concerns
       class Access < Struct
+        attribute :access, Types::String.default('dark')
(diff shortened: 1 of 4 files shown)


lib/cocina/models/collection.rb
```diff
@@ -16,6 +16,8 @@ class Collection < Struct
 
       # Subschema for access concerns
       class Access < Struct
+        attribute :access, Types::String.default('dark')
+                                        .enum('world', 'stanford', 'location-based', 'citation-only', 'dark')
```
justinlittman: I know these are specified in the maps., but what are location-based and citation-only? Are these actually supported in DOR?

jcoyne: They are presently supported in DOR. location-based means "In a reading room" and citation-only is just acknowledging we hold it in purl, but not showing the content.

PR #66, opened 2020-02-25
Add agreementId to DRO

aaron-collier:
## Why was this change made?

I believe this is part of https://github.com/sul-dlss/dor-services-app/issues/683, but could be wrong about that. 

Waiting on discussion with @jcoyne

### DIFF
--- a/lib/cocina/models/collection.rb
+++ b/lib/cocina/models/collection.rb
@@ -30,7 +30,6 @@ module Cocina
 
       # Identification sub-schema for the Collection
       class Identification < Struct
-        attribute :agreementId, Types::Strict::String.meta(omittable: true)
         attribute :catalogLinks, Types::Strict::Array.of(CatalogLink).meta(omittable: true)
       end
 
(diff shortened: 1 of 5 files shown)


lib/cocina/models/collection.rb
```diff
@@ -30,6 +30,7 @@ class Administrative < Struct
 
       # Identification sub-schema for the Collection
       class Identification < Struct
+        attribute :agreementId, Types::Strict::String.meta(omittable: true)
```
jcoyne: Looks like this should be `hasAgreement` in the structural schema: https://github.com/sul-dlss/cocina-models/blob/315b55abbb3b9980013ca1419aa51a991a9973ac/docs/maps/DRO.json#L270

aaron-collier: This pr is specifically about the identityMetadata (as posted in the https://github.com/sul-dlss/dor-services-app/issues/683 issue). Should that be udpated?

aaron-collier: @jcoyne updated. 

Do collections have agreements or just items? happy to pair if this isn't what we were thinking.

Issue #70, opened 2020-02-26
Add tests that verify that request models stay in sync with base models


justinlittman:
The request version of a model (e.g., `dro_request`) is very similar to the base model (e.g., `dro`), except, perhaps, a required field is optional. This creates the risk that a field will be added to only one of these models and not the other.

Please write a test/tests to check the attributes of these models to make sure they stay in sync (and possibly reduce duplicate tests).

ndushay:
probably:  use rspec shared_examples for tests that need to pass for both base and request models.

Somewhat related to model rename in #69 ...

Issue #69, opened 2020-02-26
[ÜBER] Rename request models


justinlittman:
To that `request` is a suffix, rather than prefix so that they sort next to the non-request models.

For example, `request_dro` becomes `dro_request`.

This needs to be done in cocina-models but also for every consumer of cocina-models, e.g. https://github.com/search?q=org%3Asul-dlss+requestdro&type=Code

ndushay:
See also #70 - not blocking, but PRs may need to be coordinate / rebased.

ndushay:
code and test code has be DRYed up;  this is no longer needed.

PR #73, opened 2020-03-03
Modules refactor

ndushay:
## Why was this change made?

first part of #69 - #70 -- refactor  XXX and RequestXXX classes to reduce code duplication.

This part is completely backwards compatible so no need to be in a feature branch.

In the next PR, I intend to put common tests into rspec shared_examples so we have less duplication in our test code.

na

## Does this change affect how this gem integrates with other services?

If so, please confirm:
- [ ] change was tested on stage    and/or
- [ ] test added to sul-dlss/infrastructure-integration-test


.rubocop_todo.yml
```diff
@@ -1,11 +1,15 @@
...
-# on 2020-02-13 12:54:33 -0500 using RuboCop version 0.79.0.
+# on 2020-03-02 18:30:42 -0800 using RuboCop version 0.74.0.
 # The point is for the user to remove these configuration records
 # one by one as the offenses are removed from the code base.
 # Note that changes in the inspected code, or installation of new
 # versions of RuboCop, may require this file to be generated again.
 
+# Offense count: 4
+Metrics/AbcSize:
+  Max: 37
```
ndushay: this is weird - it's because of something to do with `Module.included` which is an inherited method.  Rubocop didn't complain about the same thing when it was defined as part of the Class.

ndushay:
ignore the codeclimate complaints - it's actually an improvement - it reduces the duplication.

lib/cocina/models/admin_policy.rb
```diff
@@ -4,74 +4,8 @@ module Cocina
   module Models
     # An admin policy object.
     class AdminPolicy < Struct
-      include Checkable
-
-      TYPES = [
```
jcoyne: I don't think `TYPES` or `Checkable` should be part of the AdminPolicyAttributes, nor should the nested classes.

lib/cocina/models/collection_attributes.rb
```diff
@@ -0,0 +1,55 @@
...
+        attribute :hasAdminPolicy, Types::Coercible::String.optional.default(nil)
+      end
+
+      # Identification sub-schema for the Collection
+      class Identification < Struct
+        attribute :catalogLinks, Types::Strict::Array.of(CatalogLink).meta(omittable: true)
+      end
+
+      class Structural < Struct
+      end
```
jcoyne: Please remove these, they aren't part of the shared attributes.

lib/cocina/models/admin_policy_attributes.rb
```diff
@@ -0,0 +1,79 @@
...
+        # Allowing hasAdminPolicy to be omittable for now (until rolled out to consumers),
+        # but I think it's actually required for every Admin Policy
+        attribute :hasAdminPolicy, Types::Coercible::String.optional.default(nil)
+      end
+
+      class Identification < Struct
+      end
+
+      class Structural < Struct
+      end
```
jcoyne: Please remove these things, they aren't part of the shared attributes.

lib/cocina/models/dro_attributes.rb
```diff
@@ -0,0 +1,82 @@
...
+      # Identification sub-schema for the DRO
+      class Identification < Struct
+        attribute :sourceId, Types::Strict::String.meta(omittable: true)
+        attribute :catalogLinks, Types::Strict::Array.of(CatalogLink).meta(omittable: true)
+      end
+
+      # Geographic sub-schema for the DRO
+      class Geographic < Struct
+        attribute :iso19139, Types::Strict::String
+      end
```
jcoyne: Please remove these, they are not shared attributes.

lib/cocina/models/dro_attributes.rb
```diff
@@ -0,0 +1,82 @@
...
+        # but I think it's actually required for every DRO
+        obj.attribute :description, Description.optional.meta(omittable: true)
+        obj.attribute :geographic, Geographic.optional.meta(omittable: true)
+        obj.attribute(:identification, Identification.default { Identification.new })
+        obj.attribute :label, Types::Strict::String
+        obj.attribute :type, Types::String.enum(*TYPES)
+        obj.attribute :version, Types::Coercible::Integer
+      end
+
+      def image?
```
jcoyne: This is not an attribute, please remove it.

lib/cocina/models/file_attributes.rb
```diff
@@ -0,0 +1,62 @@
...
+      end
+
+      # Represents some technical aspect of the file
+      class Presentation < Struct
+        attribute :height, Types::Coercible::Integer.optional.default(nil)
+        attribute :width, Types::Coercible::Integer.optional.default(nil)
+      end
+
+      class Structural < Struct
+      end
```
jcoyne: These are not shared attributes, please restore them.

lib/cocina/models/request_file.rb
```diff
@@ -2,23 +2,11 @@
 
 module Cocina
   module Models
-    # A request to create a file.
+    # Metadata for a request to create a File.
```
jcoyne: This literally is the request itself, not just the metadata.

lib/cocina/models/collection_attributes.rb
ndushay: where do you think they should be declared, to avoid duplication?

lib/cocina/models/request_file.rb
ndushay: I will fix these comments, thanks.

lib/cocina/models/collection_attributes.rb
jcoyne: They weren’t duplicated before this PR. So the can be left where they are

ndushay: ok, thanks for the explanation

ndushay:
@jcoyne ready for re-review

PR #74, opened 2020-03-03
Group include statements in model classes

mjgiarlo:
And pull attributes up between constants and inner classes. Solicit help from Rubocop in so doing.

## Why was this change made?

To make it easier to see at a glance where included code is pulled in.

No.

## Does this change affect how this gem integrates with other services?

No.


mjgiarlo:
that's a nope. former order was needed/intended. my bad.

PR #76, opened 2020-03-04
refactor (request_)admin_policy specs to use admin_policy_shared_examples

ndushay:
## Why was this change made?

-    Ensure AdminPolicy and RequestAdminPolicy models stay in sync.

Part of #70  (akin to PR #75 for File and RequestFile models)

na

### DIFF
--- a/spec/cocina/models/admin_policy_shared_examples.rb
+++ b/spec/cocina/models/admin_policy_shared_examples.rb
@@ -1,0 +1,246 @@
+# frozen_string_literal: true
+
+require 'spec_helper'
+
+# These shared_examples are meant to be used by AdminPolicy and RequestAdminPolicy specs in
+# order to de-dup test code for all the functionality they have in common.
+# The caller must define required_properties as a hash containing
+#   the minimal required properties that must be provided to (Request)AdminPolicy.new
+RSpec.shared_examples 'it has admin_policy attributes' do
+  let(:admin_policy) { described_class.new(properties) }
+  let(:type) { Cocina::Models::Vocab.admin_policy }
+  # see block comment for info about required_properties
+  let(:properties) { required_properties }
+  let(:admin_policy_default_rights) { Cocina::Models::AdminPolicy::Administrative::DEFAULT_OBJECT_RIGHTS }
+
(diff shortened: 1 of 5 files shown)


spec/cocina/models/admin_policy_shared_examples.rb
```diff
@@ -0,0 +1,246 @@
...
+      it 'populates all attributes passed in' do
+        expect(admin_policy.administrative.hasAdminPolicy).to eq 'druid:mx123cd4567'
+        expect(admin_policy.administrative.default_object_rights).to eq '<rightsMetadata></rightsMetadata>'
+        expect(admin_policy.administrative.registration_workflow).to eq 'wasCrawlPreassemblyWF'
+
+        expect(admin_policy.description.title.first.attributes).to eq(primary: true,
+                                                                      titleFull: 'My admin_policy')
+      end
+
+      context 'with a string description title primary boolean property' do
```
justinlittman: Why are we testing this?

ndushay: to show what happens?

ndushay: it doesn't seem to be tested elsewhere, but this may not be the right place (ditto for version attribute).

Note that File.administrative boolean property DOES coerce a string to boolean

https://github.com/sul-dlss/cocina-models/blob/master/spec/cocina/models/file_shared_examples.rb#L98-L117

But Description does NOT coerce a string to boolean.

Should we accept strings or not?  Why in some places and not others?

I can pull out this out into new description_spec, and create a version_spec if you like.

ndushay: the coercion of a version value to an integer was in the existing specs, FWIW.  Not that it means it should stay there.

justinlittman: This isn't worth slowing anything down over.

ndushay: i'm gonna merge this, and work on a PR to test this stuff in a more appropriate way.

PR #77, opened 2020-03-05
description and release_tag specs for string coerced to boolean

ndushay:
## Why was this change made?

in https://github.com/sul-dlss/cocina-models/pull/76#discussion_r387955380, @justinlittman rightly questioned why I was testing that a string boolean value passed as a Description model attribute raised an error.  This PR:

- changes the Description model to have a coercible boolean value
    - I can't see the point in making it more restrictive than the other boolean types in ReleaseTag and File.Adminstrative
- has specs specific to the files declaring the models with coercible boolean values
    - just as we show the version attribute to be coercible from a string in checkable model tests
- removes the spec in admin_policy_shared_examples from PR #76 that started this whole mess.

na

### DIFF
--- a/lib/cocina/models/description.rb
+++ b/lib/cocina/models/description.rb
@@ -6,7 +6,7 @@ module Cocina
     class Description < Struct
       # Title element.  See http://sul-dlss.github.io/cocina-models/maps/Title.json
       class Title < Struct
-        attribute :primary, Types::Strict::Bool
+        attribute :primary, Types::Params::Bool # will coerce 'true' to true
         attribute :titleFull, Types::Strict::String
       end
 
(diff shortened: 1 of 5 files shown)


justinlittman:
So I would argue the other way: the coercion is completed unnecessary and if it means writing more tests, then we should get rid of it everywhere.

cc @jcoyne for thoughts.

jcoyne:
@justinlittman I think getting rid of it might be a good idea.  The data should always be correct and we shouldn't deal with conversion in this gm .

ndushay:
I would be happy to make a ticket to remove coercion from this gem for boolean and for integers;  it would not be backwards compatible so I am inclined to leave the work for another time.  Would you want to see the request classes rename before or after the coercion changes?

ndushay:
I created issue #80;  

If you want me to close this PR and instead add a PR removing the test from admin_policy_shared_examples, I am happy to do that.

ndushay:
merging this PR;  can address removal of type coercion later (it's ticketed #80)

PR #79, opened 2020-03-05
refactor (request_)collection specs to use collection_shared_examples

ndushay:
## Why was this change made?

-    Ensure Collection and RequestCollection models stay in sync.

Part of #70  (akin to PR #75, #76 for (Request)File and (Request)AdminPolicy models)

na


spec/cocina/models/request_collection_spec.rb
```diff
@@ -1,228 +1,26 @@
...
-                                                                    titleFull: 'my collection')
+  it_behaves_like 'it has collection attributes'
 
-        link = collection.identification.catalogLinks.first
-        expect(link.catalog).to eq 'symphony'
-        expect(link.catalogRecordId).to eq '44444'
-      end
-    end
+  it 'does not have Checkable model methods' do
+    item = described_class.new(required_properties)
```
jcoyne: Do we need a test for something that is not present?  There are an infinite number of things this object does not do and they aren't important to test for.

ndushay: I can take it out;  I put it in to distinguish from Collection, but will remove.

ndushay:
I have now removed `it 'does not have Checkable model methods' do` tests for RequestCollection, and added a commit to also remove the same for RequestFile and RequestAdminPolicy.

spec/cocina/models/collection_shared_examples.rb
```diff
@@ -0,0 +1,290 @@
...
+              {
+                who: 'Other Justin',
+                what: 'self',
+                date: '2017-10-20T15:42:15Z',
+                to: 'Earthworks',
+                release: 'false'
+              }
+            ]
+          },
+          description: {
```
ndushay: I removed sourceId attribute as it's not in the struct and I'm not sure we associate collection objects with a sourceId ?

PR #82, opened 2020-03-06
refactor (request_)file_set specs to use file_set_shared_examples

ndushay:
## Why was this change made?

-    Reduce duplication of test code
-    Ensure FileSet and RequestFileSet models stay in sync.

Part of #70  (akin to PR #75, #76, #77 for (Request)File, (Request)AdminPolicy, (Request)Collection models)

## Was the documentation (README, wiki) updated?

n/a

PR #83, opened 2020-03-07
refactor (request_)dro specs to use dro_shared_examples

### LINKED ISSUE #70: Add tests that verify that request models stay in sync with base models

ndushay:
## Why was this change made?

-    Ensure DRO and RequestDRO models stay in sync.

Closes #70  (akin to PR #75, #76, #77, #82 for other checkable models)

na


lib/cocina/models/dro.rb
```diff
@@ -34,19 +34,19 @@ class Embargo < Struct
...
       end
 
       # Subschema for administrative concerns
       class Administrative < Struct
-        attribute :releaseTags, Types::Strict::Array.of(ReleaseTag).meta(omittable: true).default([].freeze)
-        # Allowing hasAdminPolicy to be omittable for now (until rolled out to consumers),
+        # TODO: Allowing hasAdminPolicy to be omittable for now (until rolled out to consumers),
         # but I think it's actually required for every DRO
         attribute :hasAdminPolicy, Types::Coercible::String.optional.default(nil)
+        attribute :releaseTags, Types::Strict::Array.of(ReleaseTag).meta(omittable: true).default([].freeze)
```
ndushay: just some alphabetizing.  nothing to see here - move along.

lib/cocina/models/request_dro.rb
```diff
@@ -6,15 +6,14 @@ module Models
...
-      include DroAttributes
-
       # Structural sub-schema that contains RequestFileSet (unlike the DRO which contains FileSet)
       class Structural < Struct
         attribute :contains, Types::Strict::Array.of(RequestFileSet).meta(omittable: true)
         attribute :isMemberOf, Types::Strict::String.meta(omittable: true)
         attribute :hasMemberOrders, Types::Strict::Array.of(Sequence).meta(omittable: true)
       end
 
+      include DroAttributes
```
ndushay: I should have put this line down here before ...

spec/cocina/models/collection_shared_examples.rb
```diff
@@ -276,7 +276,7 @@
         expect(instance.administrative.hasAdminPolicy).to eq 'druid:mx123cd4567'
 
         tags = instance.administrative.releaseTags
-        expect(tags).to all(be_instance_of Cocina::Models::ReleaseTag)
+        expect(tags).to all(be_instance_of(Cocina::Models::ReleaseTag))
```
ndushay: leaving out these parens must make it somewhat ambiguous based on some failures I had when writing fileset specs ...

PR #88, opened 2020-03-11
Adds externalIdentifier as optional attribute of Request File.

justinlittman:
## Why was this change made?
External identifier is used by SDR API.

No.

### DIFF
--- a/lib/cocina/models/request_file.rb
+++ b/lib/cocina/models/request_file.rb
@@ -7,6 +7,9 @@ module Cocina
     # See http://sul-dlss.github.io/cocina-models/maps/File.json
     class RequestFile < Struct
       include FileAttributes
+      # externalIdentifier is used when submitting files to the SDR API to identify the file so that the
+      # uploaded files can be associated with the DRO.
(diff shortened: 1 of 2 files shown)


lib/cocina/models/request_file.rb
```diff
@@ -7,6 +7,7 @@ module Models
     # See http://sul-dlss.github.io/cocina-models/maps/File.json
     class RequestFile < Struct
       include FileAttributes
+      attribute :externalIdentifier, Types::Strict::String.meta(omittable: true)
```
jcoyne: Can you drop a comment here about when and how this is used, because it might be confusing.

spec/cocina/models/request_file_spec.rb
```diff
@@ -14,4 +14,18 @@
...
   it_behaves_like 'it has file attributes'
+
+  context 'when externalIdentifier provided' do
+    let(:item) { described_class.new(properties) }
+
+    let(:properties) do
+      { externalIdentifier: 'abc123' }.merge(required_properties)
+    end
+
+    # it_behaves_like 'it has file attributes'
```
jcoyne: cruft?

PR #92, opened 2020-03-19
Descriptive

jcoyne:
Note: this currently fails to generate because we don't yet support `allOf`


justinlittman:
@jcoyne Can you provide some more context / description for the ticket? It's hard to understand solely from the openapi changes.

jcoyne:
@justinlittman this is the plan for descriptive metadata that Arcadia has given me.

justinlittman:
@jcoyne I understand that. Is it described / documented anywhere? (other than the openapi).

justinlittman:
Some examples?

jcoyne:
I'll ask Arcadia to give you access.

PR #104, opened 2020-03-26
Generator spec

justinlittman:
## Why was this change made?
* To more completely test the generator.
* To stop testing the models declared in the openapi.

Yes.


justinlittman:
Yes, this is going to require some codeclimate adjustment.

README.md
```diff
@@ -26,6 +26,12 @@ exe/generator generate
 exe/generator generate_schema DRO
 ```
 
+## Testing
+
+The generator is tested via it's output when run against `openapi.yml`, viz., the Cocina model classes. Thus, `generate` should be run after any changes to `openapi.yml`.
```
mjgiarlo: I appreciate the clarification here. One nit: `s/it's/its/`

README.md
```diff
@@ -26,6 +26,12 @@ exe/generator generate
...
+
+Beyond what is necessary to test the generator, the Cocina model classes are not tested, i.e., they are assumed to as specified in `openapi.yml`.
```
mjgiarlo: I think you are missing a helper verb towards the end.

> ... are assumed to [be?] as specified in `openapi.yml`.

:grey_question:

README.md
justinlittman: Ooh, that is not a grammar error I make. Thanks for catching.

PR #108, opened 2020-04-01
Add a field for download level access

jcoyne:
## Why was this change made?

This allows us to do registration using the cocina endpoint.  This requires us to be able to set a download access level.

Ref https://github.com/sul-dlss/dor-services-app/issues/772

no

### DIFF
--- a/docs/maps/DRO.json
+++ b/docs/maps/DRO.json
@@ -84,7 +84,7 @@
         "download": {
           "description": "Download level for the DRO metadata.",
           "type": "string",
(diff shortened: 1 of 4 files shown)


jcoyne:
@mjgiarlo 1) the models were regenerated. 2) I thought we weren't going to test the outcomes as this project is tested to accurately compile openapi.yml into models.

justinlittman:
This PR should not require tests, as the outcome should be determinate from the generator (and the generator is tested).

PR #110, opened 2020-04-01
Use redoc + openapi to generate docs

jcoyne:
## Why was this change made?

Now that we use openapi, it's easier to use redoc rather than prmd.

yes.


README.md
```diff
@@ -32,24 +32,6 @@ The generator is tested via its output when run against `openapi.yml`, viz., the
...
-# Check it’s all good
-prmd verify schema.json
-
-# Build docs
-prmd doc schema.json > schema.md
-```
-
-Then check in the resulting changes to `docs/schema.json` and `docs/schema.md`
-
 ## Using this gem
```
justinlittman, line 35: Can we add a link to the openapi docs?

jcoyne: It's already in the top of the page: 
Screen Shot 2020-04-01 at 12 51 40 PM

PR #118, opened 2020-05-22
Add members to Sequence 

jcoyne:
## Why was this change made?

so that we can model virtual objects as a Sequence

See https://argo.stanford.edu/view/druid:hj097bm8879 as an example of a virtual object

### DIFF
--- a/lib/cocina/models/sequence.rb
+++ b/lib/cocina/models/sequence.rb
@@ -3,6 +3,7 @@
 module Cocina
   module Models
     class Sequence < Struct
+      attribute :members, Types::Strict::Array.of(Types::Strict::String).meta(omittable: true)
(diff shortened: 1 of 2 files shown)


jcoyne:
@mjgiarlo that's fine by me, but this is just implementing what was already designed.  See https://github.com/sul-dlss/cocina-models/blob/master/docs/maps/Sequence.json#L28-L34

mjgiarlo:
@jcoyne :speech_balloon: 
> @mjgiarlo that's fine by me, but this is just implementing what was already designed. See https://github.com/sul-dlss/cocina-models/blob/master/docs/maps/Sequence.json#L28-L34

Why am I not surprised that we already designed for this? Cool, merging. thanks!

PR #124, opened 2020-06-10
Validate the pattern of sourceId

jcoyne:
## Why was this change made?
So we don't have to write a in code validator in dor-services-app and we don't make invalid requests.

Test suite.

n/a

### DIFF
--- a/openapi.yml
+++ b/openapi.yml
@@ -842,15 +842,7 @@ components:
       additionalProperties: false
       properties:
         sourceId:
-          type: string
-          description: >
-            Unique identifier in some other system. This is because a large proportion of what is deposited in SDR,
-            historically and currently, are representations of objects that are also represented in other systems.
-            For example, digitized paper and A/V collections have physical manifestations, and those physical objects are managed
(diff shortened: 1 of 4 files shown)


justinlittman:
* Why is this pattern required?
* What are the parts of the pattern?

jcoyne:
@justinlittman I don't know why this validation is necessary, but it's the convention we currently enforce. We currently validate this at the Fedora 3 model level here: https://github.com/sul-dlss/dor-services/blob/0e35e78e41125c2269a381d7c8d131bc9e766867/lib/dor/datastreams/identity_metadata_ds.rb#L63-L65  so it makes sense to transfer this into the cocina-model.

justinlittman:
OK, then we should document in the openapi that the pattern should be `namespace:value`.

openapi.yml
```diff
@@ -977,12 +978,13 @@ components:
       properties:
         sourceId:
```
mjgiarlo, line 979: I wonder if this repetition means we ought to extract source id as its own component which can then be referred to via $ref wherever needed?

jcoyne: done

PR #125, opened 2020-06-11
Add directions for releasing the gem

jcoyne:
## Why was this change made?

We didn't yet have documentation for how to upgrade everything.

n/a

n/a

### DIFF
--- a/README.md
+++ b/README.md
@@ -32,6 +32,17 @@ The generator is tested via its output when run against `openapi.yml`, viz., the
 
 Beyond what is necessary to test the generator, the Cocina model classes are not tested, i.e., they are assumed to be as specified in `openapi.yml`.
 
+## Releasing
+
+The release process is much like any other gem. First bump the version in `lib/cocina/models/version.rb`, and commit the result. Then run:
+```
+bundle exec rake release
+```
+which pushes the gem to rubygems.org.  Next write up the release notes: https://github.com/sul-dlss/cocina-models/releases .
+
+Finally, you must release versions of [sdr-client](https://github.com/sul-dlss/sdr-client) and [dor-services-client](https://github.com/sul-dlss/dor-services-client/) pinned to this version because [Argo](https://github.com/sul-dlss/argo) depends on both of those. When [dor-services-app](https://github.com/sul-dlss/dor-services-app) is updated to use the new models (via the auto-update script), the clients must be updated at the same time or there is risk of models produced by dor-services-app not being acceptable to the clients.
+
+
 ## Using this gem
 
 If you are using this gem in an application that has an API that accepts Cocina models (e.g., SDR API, Dor-Services-App), make sure that the `openapi.yml` for the application includes the schemas that match the schemas in this `openapi.yml`.


README.md
```diff
@@ -32,6 +32,18 @@ The generator is tested via its output when run against `openapi.yml`, viz., the
...
 
+## Releasing
+
+The release process is much like any other gem. First bump the version in `lib/cocina/models/version.rb`, and commit the result. Then run:
+```
+bundle exec rake release
+```
+which pushes the gem to rubygems.org.  Next write up the release notes: https://github.com/sul-dlss/cocina-models/releases .
+
+Finally, you must release versions of [sdr-api](https://github.com/sul-dlss/sdr-api) and [dor-services-client](https://github.com/sul-dlss/dor-services-client/) pinned to this version because [Argo](https://github.com/sul-dlss/argo) depends on both of those. When [dor-services-app](https://github.com/sul-dlss/dor-services-app) is update to use the new models, the clients must be updated at the same time or there is risk of models produced by dor-services-app not being acceptable to the clients.
```
justinlittman, line 43: sdr-api client and all consumers of client (viz, Google Books)?

jcoyne: I was omitting "all consumers of client" because the auto-update script is handling that.  The autoupdate script breaks in the case when we don't update sdr-client and dor-services-client both at the same time and I offered one example how.

justinlittman: Doesn't sdr-client need to be updated and released, as it is pinned to a particular cocina-models version: https://github.com/sul-dlss/dor-services-client/blob/master/dor-services-client.gemspec#L26?

jcoyne: yes, I typoed sdr-client as sdr-api here.

README.md
```diff
@@ -32,6 +32,17 @@ The generator is tested via its output when run against `openapi.yml`, viz., the
...
 
+## Releasing
+
+The release process is much like any other gem. First bump the version in `lib/cocina/models/version.rb`, and commit the result. Then run:
+```
+bundle exec rake release
+```
+which pushes the gem to rubygems.org.  Next write up the release notes: https://github.com/sul-dlss/cocina-models/releases .
+
+Finally, you must release versions of [sdr-client](https://github.com/sul-dlss/sdr-client) and [dor-services-client](https://github.com/sul-dlss/dor-services-client/) pinned to this version because [Argo](https://github.com/sul-dlss/argo) depends on both of those. When [dor-services-app](https://github.com/sul-dlss/dor-services-app) is updated to use the new models (via the auto-update script), the clients must be updated at the same time or there is risk of models produced by dor-services-app not being acceptable to the clients.
```
ndushay, line 43: Might I suggest a little formatting for skimmability:

```
Finally, you MUST release versions of **[sdr-client](https://github.com/sul-dlss/sdr-client)** and **[dor-services-client](https://github.com/sul-dlss/dor-services-client/)** pinned to the newly released cocina-models version because [Argo](https://github.com/sul-dlss/argo) depends on both of those. 

This also addresses:

When [dor-services-app](https://github.com/sul-dlss/dor-services-app) is updated to use the newly released models (via the auto-update script), the dor-services-clients used by other apps must be updated at the same time or there is risk of models produced by dor-services-app not being acceptable to the clients.

Thus it is vital that the sdr-client and dor-services-client gems are updated and released at the same time, and that all consumers, direct and indirect, are updated at the same time.
```

PR #135, opened 2020-07-10
Generate without depending on rubocop for formatting

jcoyne:
## Why was this change made?

Less noise in the output

Test suite

n/a

### DIFF
--- a/lib/cocina/generator/schema.rb
+++ b/lib/cocina/generator/schema.rb
@@ -15,12 +15,7 @@ module Cocina
           module Cocina
             module Models
               class #{name} < Struct
-
-                #{types}
-
-                #{model_attributes}
-
-                #{validate}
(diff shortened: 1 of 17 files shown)


jcoyne:
@justinlittman it was just a lot of noise and made the output less useful.   How did you get burned before?

justinlittman:
I was spending a lot of time fiddling with spaces trying to get the layout right. I'd argue that worrying about formatting in code makes the code harder to understand / change, albeit at the expense of more noisy output.

To be clear, this represents my opinion, not a hard change requirement on this ticket.

mjgiarlo:
@jcoyne @justinlittman How about this alternative approach? https://github.com/sul-dlss/cocina-models/pull/161

mjgiarlo:
@jcoyne shall we close this now that #161 is merged? thanks for working to make our tools kinder to us

PR #136, opened 2020-08-18
Add controlledDigitalLending setting for DROAccess

jcoyne:
## Why was this change made?
So that the cocina model can indicate which objects have CDL.

Test suite.


openapi.yml
```diff
@@ -567,6 +567,10 @@ components:
             - 'citation-only'
             - 'dark'
           default: 'dark'
+        controlledDigitalLending:
```
peetucket, line 570: where does the camel casing happen?   trying to understand how this maps to the work in dor-services-app

lib/cocina/models/dro_access.rb
```diff
@@ -4,6 +4,8 @@ module Cocina
   module Models
     class DROAccess < Struct
       attribute :access, Types::Strict::String.default('dark').enum('world', 'stanford', 'location-based', 'citation-only', 'dark').meta(omittable: true)
```
peetucket, line 6: do we need a new value here for cdl?

jcoyne: No, because CDL is separate from the access value.  Which would continue to be `stanford`

openapi.yml
jcoyne: I just typed it in this way.

peetucket:
Looks a small rubocop issue

jcoyne:
@peetucket https://github.com/sul-dlss/cocina-models/pull/138 should fix the rubocop

PR #141, opened 2020-08-25
add cdl file access support

### LINKED ISSUE #140: Handle file level rights for controlled digital lending
Controlled digital lending is adding file level access rights and needs to be represented in controlled digital lending: https://github.com/sul-dlss/dor-rights-auth/pull/47

peetucket:
## Why was this change made?

Fix #140

### DIFF
--- a/lib/cocina/models/access.rb
+++ b/lib/cocina/models/access.rb
@@ -4,7 +4,7 @@ module Cocina
   module Models
     class Access < Struct
       # Access level
-      attribute :access, Types::Strict::String.default('dark').enum('world', 'stanford', 'location-based', 'citation-only', 'dark').meta(omittable: true)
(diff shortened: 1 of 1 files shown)


lib/cocina/models/access.rb
```diff
@@ -4,7 +4,7 @@ module Cocina
   module Models
     class Access < Struct
       # Access level
-      attribute :access, Types::Strict::String.default('dark').enum('world', 'stanford', 'location-based', 'citation-only', 'dark').meta(omittable: true)
+      attribute :access, Types::Strict::String.default('dark').enum('world', 'stanford', 'location-based', 'citation-only', 'dark', 'cdl').meta(omittable: true)
```
jcoyne, line 7: This should be a separate property as we do for DRO access

jcoyne: We make changes to this project by updating openapi.yml and running the generator

peetucket:
Closed in favor of #142

PR #146, opened 2020-09-08
Allow the generator to handle cases where the type is anyOf: [integer…

### LINKED ISSUE #145: Can't generate DescriptiveBasicValue
This PR added a string or integer value: https://github.com/sul-dlss/cocina-models/pull/144/files

jcoyne:
…, string]

## Why was this change made?
Fixes #145


justinlittman (changes requested):
Test for change to the generator?

justinlittman (approved):

PR #151, opened 2020-09-10
Update descriptive model

arcadiafalcone:
## Why was this change made?

To make model compliant with design principles and add a missing required property

swagger.io

cocina-descriptive-metadata/description-cocina-schema_v4-3.json

### DIFF
--- a/openapi.yml
+++ b/openapi.yml
@@ -472,11 +472,11 @@ components:
               allOf:
                 - $ref: "#/components/schemas/Standard"
                 - type: object
-                properties:
-                  valueScript:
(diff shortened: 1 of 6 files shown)


openapi.yml
```diff
@@ -1220,17 +1230,6 @@ components:
           type: array
           items:
             $ref: "#/components/schemas/DescriptiveValue"
-    SourceId:
```
jcoyne, line 1223: I would like these schemas to be in alphabetical order. I can fix this though.

mjgiarlo (approved):

PR #152, opened 2020-09-10
Extract a DescriptiveValueLanguage schema

jcoyne:
## Why was this change made?
The generator can't cope with allOf for an attribute


jcoyne:
Arcadia signed off on this in Slack.

PR #159, opened 2020-09-29
Remove download access from Access schema

jcoyne:
## Why was this change made?
The Access schema is used only by collections. DROs and Files have their own Access schema which supports download access


mjgiarlo (approved):
:ok_hand: 

cc: @ndushay, since you're touching collection access stuff in your hydrus work

Issue #154, opened 2020-09-16
Description: allow status to have integer or string value


arcadiafalcone:

arcadiafalcone:
Blocked by https://github.com/interagent/committee/issues/286

ndushay:
See PR https://github.com/sul-dlss/cocina-models/pull/183 ?

arcadiafalcone:
Closed by #183

PR #167, opened 2020-10-14
Update Description

arcadiafalcone:
## Why was this change made?

Updates to Description schema

Online Swagger editor

cocina-descriptive-metadata

### DIFF
--- a/openapi.yml
+++ b/openapi.yml
@@ -986,6 +986,9 @@ components:
           format: uri
         access:
           $ref: "#/components/schemas/DescriptiveAccessMetadata"
+        version:
+          description: The version of the related resource.
+          type: string
     ReleaseTag:
       description: A tag that indicates the item or collection should be released.
       type: object
--- a/openapi.yml
+++ b/openapi.yml
@@ -301,7 +301,7 @@ components:
             $ref: "#/components/schemas/Event"
         form:
           description: Characteristics of the resource's physical, digital, and intellectual
-            form and genre.
+            form and genre, and of its process of creation.
           type: array
           items:
             $ref: "#/components/schemas/DescriptiveValue"
--- a/openapi.yml
+++ b/openapi.yml
@@ -436,7 +436,7 @@ components:
               description: Status of the descriptive element value relative to other instances
                 of the element.
               type: string
-              # type: integer
+              # type: integer (uncomment when issue #154 is resolved)
             code:
               description: Code value of the descriptive element.
               type: string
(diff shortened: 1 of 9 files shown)


openapi.yml
```diff
@@ -435,6 +436,7 @@ components:
               description: Status of the descriptive element value relative to other instances
                 of the element.
               type: string
+              # type: integer
```
justinlittman, line 439: Cruft?

openapi.yml
```diff
@@ -950,6 +972,9 @@ components:
           type: array
           items:
             $ref: "#/components/schemas/DescriptiveValue"
+        standard:
+          #description: Descriptive or content standards to which the value conforms.
```
justinlittman, line 976: Cruft?

openapi.yml
```diff
@@ -1263,8 +1291,8 @@ components:
           type: array
           items:
             $ref: "#/components/schemas/DescriptiveValue"
-        source:
-          $ref: "#/components/schemas/Source"
         version:
-          description: The version of the standard or encoding.
+          description: The version of the value standard.
```
justinlittman, line 1295: What is a value standard?

justinlittman (changes requested):

openapi.yml
jcoyne: No, you can't provide description with a $ref. So this is just an annotation for humans.

justinlittman: OK. Recommend `# description` to make it seem a bit less crufty but not a big deal.

openapi.yml
arcadiafalcone: See #154 . It's in the conceptual model but can't yet be implemented due to a bug elsewhere.

openapi.yml
arcadiafalcone: Value standard is about how you derive and construct the metadata values. For example, MARC says the title goes in 245$a, but it's the content standard RDA that has rules about where you get the title from (cover? title page? spine?), how you transcribe it, what kinds of title variations should be included, how to handle a typo on the resource, etc. Other examples of library value standards are DACS (archives) and the various RBMS standards (rare books and manuscripts).

openapi.yml
justinlittman: Can you add that to the comment?

openapi.yml
justinlittman: Can you add `e.g., RDA`? At least for me, that would make it much clearer.

justinlittman (approved):

PR #168, opened 2020-10-16
Separate language to its own model

jcoyne:
## Why was this change made?
Previously it was not accepting script as a valid property. Possibly becase it was doing allOf with members that specified additionalProperties: false

Ref https://github.com/sul-dlss/dor-services-app/issues/1157


mjgiarlo (approved):

mjgiarlo:
@jcoyne feel free to merge if we don't care about the coverage drop

PR #173, opened 2020-10-21
Set some additionalProperties to true.

justinlittman:
## Why was this change made?
There are some cases where setting additionalProperties to false causes committee problems. When the openapi is copied from cocina models to dor services app or SDR API, it is REAL easy to forget to make those changes (and the errors come out sideways). By placing these changes directly in this openapi it will avoid this problem.

NA

openapi

### DIFF
--- a/openapi.yml
+++ b/openapi.yml
@@ -137,7 +137,8 @@ components:
         - hasAdminPolicy
     AdminPolicy:
       type: object
-      additionalProperties: false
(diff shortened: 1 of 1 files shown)


jcoyne:
@justinlittman I think it's important to keep these `additionalProperties: false` in here to keep the data clean.  A user should know if they had a typo in their request for example.    Can we address the particular issues (maybe by removing oneOf, or allOf assertions)?

PR #177, opened 2020-10-22
Update classification property to subject

aaron-collier:
## Why was this change made?

fixes https://github.com/sul-dlss/dor-services-app/issues/1231 by replacing the individual properties with the required subject property.

Unit

N/A

### DIFF
--- a/lib/cocina/models/classification.rb
+++ b/lib/cocina/models/classification.rb
@@ -3,9 +3,7 @@
 module Cocina
   module Models
     class Classification < Struct
-      attribute :authority, Types::Strict::String.meta(omittable: true)
-      attribute :displayLabel, Types::Strict::String.meta(omittable: true)
-      attribute :edition, Types::Strict::String.meta(omittable: true)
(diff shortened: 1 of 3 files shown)


mjgiarlo (approved):

lib/cocina/models/classification.rb
```diff
@@ -3,9 +3,7 @@
 module Cocina
   module Models
     class Classification < Struct
-      attribute :authority, Types::Strict::String.meta(omittable: true)
-      attribute :displayLabel, Types::Strict::String.meta(omittable: true)
-      attribute :edition, Types::Strict::String.meta(omittable: true)
+      attribute :subject, Types::Strict::Array.of(DescriptiveValue).meta(omittable: true)
```
jcoyne, line 6: I don't think this is right. See https://github.com/sul-dlss-labs/cocina-descriptive-metadata/blob/master/mods_cocina_mappings/mods_to_cocina_classification.txt

PR #183, opened 2020-11-09
Make Description status string-only and add order property for integer values

arcadiafalcone:
…r values to Contributor and RelatedResource

## Why was this change made?
To get around bug preventing status from having both string and integer values, and to support H2 descriptive metadata mapping.

Online swagger.io validator

cocina-descriptive-metadata JSON schemas

### DIFF
--- a/openapi.yml
+++ b/openapi.yml
@@ -261,7 +261,6 @@ components:
           description: Status of the contributor relative to other parallel contributors
             (e.g. the primary author among a group of contributors).
           type: string
-          # type: integer (uncomment when issue #154 resolved)
         role:
           description: Relationships of the contributor to the resource or to an event
             in its history.
@@ -278,6 +277,9 @@ components:
           type: array
           items:
             $ref: "#/components/schemas/DescriptiveValue"
+        order:
+          description: Place of contributor in order of authorship for resource.
+          type: integer
     Description:
       type: object
       additionalProperties: false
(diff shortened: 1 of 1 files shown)


openapi.yml
```diff
@@ -899,63 +881,17 @@ components:
...
-          $ref: "#/components/schemas/DescriptiveValueLanguage"
+      allOf:
+        - $ref: '#/components/schemas/DescriptiveValue'
+        - type: object
+          # additionalProperties breaks the validator, presumably because it can't
+          # conform to other schemas (allOf) and not have additionalProperties
+          # additionalProperties: false
+          properties:
+            script:
+              $ref: '#/components/schemas/DescriptiveValue'
```
ndushay, line 892: I don't understand this change, and maybe that's ok.  The comments above are confusing me further.   it seems like we've gone from a list of properties to a script property that is nested?

ndushay (commented):
I'm leaving a comment because i don't want to block ... but the description of this PR and what is in it ... confuse me.  what is here is not backwards compatible -- we are removing stuff.  And maybe that's ok, but I just don't have info as to why.

openapi.yml
```diff
@@ -441,7 +443,6 @@ components:
               description: Status of the descriptive element value relative to other instances
                 of the element.
               type: string
-              # type: integer (uncomment when issue #154 is resolved)
```
ndushay, line 444: does this one need a an "order" property added?

ndushay (changes requested):
one question

openapi.yml
arcadiafalcone: No, there hasn't been a use case for that yet, so I decided to leave it out for now.

ndushay (approved):

PR #194, opened 2020-12-11
pr-template gets note about propagating openapi.yml changes

ndushay:
**NOTE:  Changes to openapi.yml require generating models - see README.**

## Why was this change made?

For the clueless among us, having the pr template ALSO say that we need to update openapi.yml in other apps is helpful.


justinlittman (approved):

jcoyne (changes requested):
I don't think this belongs in a PR template, because this action has to happen when the version is released and rolled out to the consumers.

ndushay:
It's for dodoes like me -- I've now merged openapi.yml changes twice without thinking about this other work that needs doing.  Feel free to change the wording or whatever.

PR #198, opened 2021-01-06
Change adminMetadata.standard to adminMetadata.metadataStandard to ac…

arcadiafalcone:
…commodate records with multiple standards

## Why was this change made?
To address https://github.com/sul-dlss-labs/cocina-descriptive-metadata/issues/232

Swagger.io

cocina-descriptive-metadata, stanford-library-metadata


jcoyne (changes requested):
Can you generate the models to go with this change?

arcadiafalcone:
@jcoyne New model has been generated.

jcoyne (approved):

PR #201, opened 2021-01-09
Add a field so that AdminPolicy can communicate about which collectio…

jcoyne:
…ns are authorized for registration

### DIFF
--- a/openapi.yml
+++ b/openapi.yml
@@ -169,6 +169,11 @@ components:
           default: <?xml version="1.0" encoding="UTF-8"?><rightsMetadata><access type="discover"><machine><world/></machine></access><access type="read"><machine><world/></machine></access><use><human type="useAndReproduction"/><human type="creativeCommons"/><machine type="creativeCommons" uri=""/><human type="openDataCommons"/><machine type="openDataCommons" uri=""/></use><copyright><human/></copyright></rightsMetadata>
         registrationWorkflow:
           type: string
+        collectionsForRegistration:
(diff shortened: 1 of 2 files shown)


openapi.yml
```diff
@@ -169,6 +169,11 @@ components:
           default: <?xml version="1.0" encoding="UTF-8"?><rightsMetadata><access type="discover"><machine><world/></machine></access><access type="read"><machine><world/></machine></access><use><human type="useAndReproduction"/><human type="creativeCommons"/><machine type="creativeCommons" uri=""/><human type="openDataCommons"/><machine type="openDataCommons" uri=""/></use><copyright><human/></copyright></rightsMetadata>
         registrationWorkflow:
           type: string
+        collections_for_registration:
+          description: When you register an item with this admin policy, these are the collections that are available.
```
justinlittman, line 173: Is this simply an Argo configuration? Or is this an enforced restriction? What about after registration?

jcoyne: Currently it serves as Argo configuration, however it is stored in the models in dor-services.  One could make a convincing argument that is configuration for Argo, but as we introduce sdr-api, we likely need it there as well.  This is just a naive approach to decoupling Argo from Fedora 3.  See https://github.com/sul-dlss/argo/issues/2354

mjgiarlo: Fine by me. Would you consider renaming this to be consistent with its siblings? Perhaps `registrationCollections` or `collectionsForRegistration` ?

justinlittman (approved):

PR #210, opened 2021-02-26
Update model for originInfo/event use cases

arcadiafalcone:
**NOTE:  Changes to openapi.yml require updating openapi.yml for sdr-api and dor-services-app and generating models - see README.**

## Why was this change made?
Facilitates originInfo/event roundtripping

## How was this change tested?
generator

## Which documentation and/or configurations were updated?
none

PR #214, opened 2021-03-03
Update openapi.yml for Event mapping use cases

ndushay:

- I have branches with openapi.yml updated for both dor-services-app and sdr-api;  waiting for this the PR to be updated and for the gem release to move those forward.
- I will coordinate with FR and JCoyne re: updating dor-services-client and sdr-client gems, and the remaining PRs.  
- I believe the changes from THIS PR would only require updates to dor-services-app, as all the event handling is being done there.  🤷‍♀️ 
- I believe there are other changes since last gem release having to do with H2, which @jcoyne knows about.

## Why was this change made?

Event modeling changes to enable MODS originInfo mapping that will roundtrip cleanly.

This is a redo of Arcadia's PR #210, but rebased on main after PRs #212 and #213 were merged into main.

circleci, mainly ... the openapi validator and rspec tests.

### DIFF
--- a/openapi.yml
+++ b/openapi.yml
@@ -572,6 +572,47 @@ components:
           type: array
           items:
             $ref: "#/components/schemas/DescriptiveValue"
+    DescriptiveParallelEvent:
+      description: Value model for multiple representations of information about the same event (e.g. in different languages).
+      type: object
+      additionalProperties: false
+      allOf:
+        - $ref: "#/components/schemas/DescriptiveValueLanguage"
+        - $ref: "#/components/schemas/DescriptiveStructuredValue"
+        - type: object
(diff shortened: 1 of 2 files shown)


justinlittman (approved):

jcoyne (approved):

openapi.yml
```diff
@@ -794,7 +835,6 @@ components:
       additionalProperties: false
       allOf:
         - $ref: "#/components/schemas/DescriptiveStructuredValue"
```
arcadiafalcone, line 837: I realized a $ref for DescriptiveValueLanguage should be added to Event after all, to correspond to originInfo attributes.

PR #220, opened 2021-03-04
Fix valueLanguage in event models

arcadiafalcone:

## Why was this change made?
- Add missing valueLanguage
- move Title to the correct place alphabetically in openapi.yml

CircleCI

n/a


ndushay (approved):
I squashed some commits, but that's it, so I'm approving and merging this.

PR #233, opened 2021-03-16
Add barcode to identification schema

aaron-collier:

## Why was this change made?

This allows for capturing other identifiers (barcode, catkey, etc.) without adding individual fields.

Refs: https://github.com/sul-dlss/dor_indexing_app/issues/509


jcoyne (changes requested):
This seems like a departure from the catkey and sourceId where we explictly modeled as a field.  Is there a reason for that? If we model as specific keys then we can place constraints on them so we can catch data errors.

openapi.yml
```diff
@@ -1180,6 +1184,17 @@ components:
         width:
           description: Width in pixels
           type: integer
+    OtherId:
+      description: Other identifiers on DRO objects
+      type: object
+      additionalProperties: false
+      properties:
+        type:
+          description: The type of an other identifier (i.e. catkey, barcode)
```
jcoyne, line 1193: Catkey is already in the model, so it is confusing to mention it here.

jcoyne (changes requested):

aaron-collier:
@jcoyne being able to constrain this is a good point, I'll do a bit of analysis to see what else may need to be captured in other ids, and come back to this.

aaron-collier:
I'm not sure we'll be able to effectively validate barcodes, from libsys:

```
No, we don't have any validation on barcodes in Symphony.  Partly because they CAN take multiple forms.  The one you describe is auto-generated by the system, and consists of the <ckey>-<callseg+3-digit-copynum>

There are also different forms for the barcodes GSB uses.

For SUL, for real barcodes, they should start with 36105 and be 14 digits.  SHOULD.  There are typos all the time.
```

jcoyne:
@aaron-collier why not go with those two forms we know and then we can expand the definition as needed.

aaron-collier:
@jcoyne updated.

andrewjbtw:
@aaron-collier we definitely can't assume the 36105 prefix. We have business library barcodes that start with a different prefix. Ex: https://argo.stanford.edu/view/druid:cp590xz6533

andrewjbtw:
Sara's comment here suggests business barcodes will always start `2050`: https://github.com/sul-dlss/argo/issues/2315#issuecomment-743479323

openapi.yml
```diff
@@ -248,6 +248,10 @@ components:
           type: array
           items:
             $ref: "#/components/schemas/DescriptiveBasicValue"
+    Barcode:
```
jcoyne, line 251: Is it possible to model the three known types of barcodes separately, then we'd know something of their provenance.  We could have `StandardBarcode: '^36105[0-9]{9}$'`, `BusinessBarcode: '^2050[0-9]{10}$'`, `CatkeyBarcode: '^[0-9]{7}-[0-9]{4}$'`

then:
```yaml
  barcode:
    oneOf:
      - $ref: '#/components/schemas/StandardBarcode'
      - $ref: '#/components/schemas/BusinessBarcode'
      - $ref: '#/components/schemas/CatkeyBarcode'
```

mjgiarlo:
Not to pull this back too far, but do we have a set of functional requirements around barcodes that could help ground this work? If I had a greater understanding of how barcodes are used in Argo and upstream (discovery / delivery stack), and why, that would let me grok what changes are needed. It'd be good to seek and provide this sort of information for cocina-models changes going forward, if not too much of a drag.

openapi.yml
aaron-collier: I like this.

aaron-collier:
At a minimum, I believe we must index the barcode to support CDL, however @andrewjbtw may have more to say about further requirements.

mjgiarlo:
After talking with folks a bit about this, here's what I learned: yes, the barcode is used for CDL. It must be in the public XML in PURL (`<otherId name="barcode">...</otherId>`), so it is reasonable to model this in cocina-models.

andrewjbtw:
@mjgiarlo more background https://github.com/sul-dlss/argo/issues/2315

jcoyne (approved):

PR #235, opened 2021-03-17
Add properties for modeling the admin policy defaults

jcoyne:

## Why was this change made?

Currently we're just using an unvalidated chunk of XML to convey these values, but we should migrated to explicit values.

### DIFF
--- a/lib/cocina/models/admin_policy_administrative.rb
+++ b/lib/cocina/models/admin_policy_administrative.rb
@@ -3,7 +3,9 @@
 module Cocina
   module Models
     class AdminPolicyAdministrative < Struct
+      # This is an XML expression of the default access (see defaultAccess)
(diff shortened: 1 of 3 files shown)


andrewjbtw:
I think there are fields in defaultObjectRights that we are not mapping in the rightsMetadata yet. See https://github.com/sul-dlss/cocina-models/issues/236

There's only a small number of APOs, so we should be able to check all of the defaultObjectRights that currently exist to see if any have other data not being mapped.

mjgiarlo (approved):

mjgiarlo (changes requested):
Do we need to do more analysis re: @andrewjbtw's question before we merge this?

aaron-collier:
@mjgiarlo I think we do, and likely should have a specific meeting like we did about version.

justinlittman (approved):

justinlittman:
Per discussion at today's standup, OK to proceed with this PR.

PR #245, opened 2021-03-26
Mark fields that are not required as optional

### LINKED ISSUE #244: String values that are omittable should also be optional
Otherwise it's very hard to delete an existing value by doing an update.

jcoyne:
## Why was this change made?

this gives us a way to do an update where we clear a value without having to have all the values of a schema

Fixes #244

### DIFF
--- a/lib/cocina/generator/schema_base.rb
+++ b/lib/cocina/generator/schema_base.rb
@@ -21,10 +21,20 @@ module Cocina
         key || schema_doc.name
       end
 
+      # Allows non-required values to not be provided. This allows smaller
+      # requests as not every field needs to be present.
       def omittable
         return '' if required
 
         '.meta(omittable: true)'
+      end
+
+      # Allows non-required values to be set to nil. This is useful when doing
+      # an update and you want to clear out a value.
+      def optional
+        return '' if required
+
+        '.optional'
       end
 
       def quote(item)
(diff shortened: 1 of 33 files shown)


justinlittman (changes requested):
Can you add a spec for generator?

jcoyne:
@justinlittman you bet! I added one.

lib/cocina/generator/schema_base.rb
```diff
@@ -21,12 +21,20 @@ def name
         key || schema_doc.name
       end
 
+      # Allows values to be not be provided. This allows smaller requests as not every field needs to be provided.
```
ndushay, line 24: "to be not be provided"?  change to "not to be provided"?

lib/cocina/generator/schema_base.rb
```diff
@@ -21,12 +21,20 @@ def name
...
       def omittable
         return '' if required
 
         '.meta(omittable: true)'
       end
 
+      # Allows values to be set to nil. This is useful when doing an update and you want to clear out a value.
```
ndushay, line 31: "Allows non-required values to be set to nil" ???

ndushay (changes requested):
Couple of comment quibbles;  o.w. looks okay to merge but I think 

Since this change impacts every model, I think it warrants running the integration tests before rolling it to prod.  (that is, after gem is released with this change, deploy to qa and run integration tests).

jcoyne:
@ndushay thanks for the constructive suggestions. Changes made.

ndushay (approved):
approving as my comment quibbles were addressed;  I also don't want to block this while I'm out with the 2nd jab reaction.  I stand by my statement about testing this via integration tests before rolling it to prod.

PR #250, opened 2021-03-31
Only fields marked as nullable can be optional

jcoyne:
## Why was this change made?

Because although our models were accepting null, the validator, which is just validating the OAS 3.0 spec was saying it wasn't valid to send null for the properties we were trying to clear.


mjgiarlo (commented):
Looks like a whooooole lotta `.optional` is gone now. Intended?

jcoyne:
@mjgiarlo yes, this reverts the #245 changes and tracks nullable rather than required.

mjgiarlo (approved):

mjgiarlo (approved):

PR #251, opened 2021-04-02
Expand cocina-models documentation to make release process clearer

### LINKED ISSUE #247: Clarify README about orchestration of changes when models change
I.e.:

1. Update cocina-models
2. Update clients
3. Update dsa & sdr-api
4. Update other codebases as necessary depending on whether bw-compat or not, etc.

mjgiarlo:
Fixes #247

## Why was this change made?

Increased clarity for the team

N/A

README

### DIFF
--- a/README.md
+++ b/README.md
@@ -34,44 +34,53 @@ Beyond what is necessary to test the generator, the Cocina model classes are not
 
 ## Releasing
 
+### Step 0: Share intent to change the models
+
+Send a note to `#dlss-infra-chg-mgmt` on Slack to let people know what is changing and when.
+
+### Step 1: Cut the release
+
 The release process is much like any other gem. First bump the version in `lib/cocina/models/version.rb`, and commit the result. Then run:
 ```
 bundle exec rake release
 ```
 which pushes the gem to rubygems.org.  Next write up the release notes: https://github.com/sul-dlss/cocina-models/releases .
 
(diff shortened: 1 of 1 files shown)


README.md
```diff
@@ -34,44 +34,53 @@ Beyond what is necessary to test the generator, the Cocina model classes are not
...
+
+### Step 1: Cut the release
+
 The release process is much like any other gem. First bump the version in `lib/cocina/models/version.rb`, and commit the result. Then run:
 ```
 bundle exec rake release
 ```
 which pushes the gem to rubygems.org.  Next write up the release notes: https://github.com/sul-dlss/cocina-models/releases .
 
-Finally, you must release versions of [sdr-client](https://github.com/sul-dlss/sdr-client) and [dor-services-client](https://github.com/sul-dlss/dor-services-client/) pinned to this version because [Argo](https://github.com/sul-dlss/argo) depends on both of those. When [dor-services-app](https://github.com/sul-dlss/dor-services-app) is updated to use the new models (via the auto-update script), the clients must be updated at the same time or there is risk of models produced by dor-services-app not being acceptable to the clients.
```
mjgiarlo, line 43: None of the text in the README was actually removed. It was just shifted around a bit to hopefully communicate the organization and order of information more clearly and more effectively.

README.md
```diff
@@ -34,44 +34,53 @@ Beyond what is necessary to test the generator, the Cocina model classes are not
...
+### Step 2: Update client gems coupled to the models
 
-### Communicate
+Next, you should release versions of [sdr-client](https://github.com/sul-dlss/sdr-client) and [dor-services-client](https://github.com/sul-dlss/dor-services-client/) pinned to this version because applications such as [Argo](https://github.com/sul-dlss/argo) depend on both of these gems using the same models.
```
ndushay, line 51: do you want to say anything about needing to do this before updating APIs due to roundtrip validation in dor-services-app?

ndushay (approved):
approved and dropped a question

README.md
mjgiarlo: I'd rather not get into the reasons why, if I could avoid it!

PR #254, opened 2021-04-15
Adds additional access attributes for embargoes and moves some access…

### LINKED ISSUE #253: embargo release access rights values should support all values in the Argo rights menu
from https://argo.stanford.edu/catalog?f%5BobjectType_ssim%5D%5B%5D=item

as the issue title says:  embargo release access rights values should support all values in the Argo rights menu:

image

Per @andrewjbtw, the cocina embargo model should support all these settings.  With a bit of a dither:

> I should add that arguably, an embargo should only support increasing rather than limiting access but
> 
> - it may be simpler to show the same rights menu here as elsewhere instead of customizing just for embargoes
> - only citation-only and dark would be excluded if we excluded anything
> - the idea was suggested for digital borrowing that items could be set to "expire" at the end of the quarter by automatically changing to citation-only or dark via embargo release. We have not been asked to implement this but it also has not been categorically ruled out.

(so consider the above, think about whether this restriction should happen HERE or in Argo (see sul-dlss/argo/issues/2674) and possibly double check with andrew before setting a PR for this ticket as "ready for review."  My gut reaction is that the restriction should happen in Argo and/or dor-services-app)

Currently we have the possibilities as https://github.com/sul-dlss/cocina-models/blob/main/lib/cocina/models/embargo.rb#L10 

image

but note that the nuances are expressed differently in https://github.com/sul-dlss/cocina-models/blob/main/lib/cocina/models/access.rb with separate attributes for `readLocation`:

image

And that the `download` attribute is specified in 

https://github.com/sul-dlss/cocina-models/blob/main/lib/cocina/models/admin_policy_default_access.rb
https://github.com/sul-dlss/cocina-models/blob/main/lib/cocina/models/dro_access.rb
https://github.com/sul-dlss/cocina-models/blob/main/lib/cocina/models/file_access.rb

justinlittman:
… attributes into own schema.

closes #253

## Why was this change made?

NA

openapi

### DIFF
--- a/lib/cocina/models/access.rb
+++ b/lib/cocina/models/access.rb
@@ -3,13 +3,13 @@
 module Cocina
   module Models
     class Access < Struct
-      # Access level
+      # Access level that applies when embargo expires.
       attribute :access, Types::Strict::String.default('dark').enum('world', 'stanford', 'location-based', 'citation-only', 'dark').meta(omittable: true)
       # The human readable copyright statement that applies
       # example: Copyright World Trade Organization
       attribute :copyright, Types::Strict::String.meta(omittable: true)
       # If access is "location-based", which location should have access.
-      attribute :readLocation, Types::Strict::String.enum('spec', 'music', 'ars', 'art', 'hoover', 'm&m').meta(omittable: true)
+      attribute :readLocation, Types::Strict::String.optional.enum('spec', 'music', 'ars', 'art', 'hoover', 'm&m').meta(omittable: true)
       # The human readable use and reproduction statement that applies
       # example: Property rights reside with the repository. Literary rights reside with the creators of the documents or their heirs. To obtain permission to publish or reproduce, please contact the Public Services Librarian of the Dept. of Special Collections (http://library.stanford.edu/spc).
       attribute :useAndReproductionStatement, Types::Strict::String.meta(omittable: true)
--- a/lib/cocina/models/access_read_location.rb
+++ b/lib/cocina/models/access_read_location.rb
@@ -1,0 +1,9 @@
+# frozen_string_literal: true
+
+module Cocina
+  module Models
+    AccessReadLocation = Types::String.constrained(
+      format: //i
+    )
+  end
+end
--- a/lib/cocina/models/dro_access.rb
+++ b/lib/cocina/models/dro_access.rb
@@ -3,6 +3,7 @@
 module Cocina
   module Models
     class DROAccess < Struct
+      # Access level that applies when embargo expires.
       attribute :access, Types::Strict::String.default('dark').enum('world', 'stanford', 'location-based', 'citation-only', 'dark').meta(omittable: true)
       # Available for controlled digital lending.
       attribute :controlledDigitalLending, Types::Strict::Bool.meta(omittable: true)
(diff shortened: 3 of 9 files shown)


openapi.yml
```diff
@@ -99,36 +99,54 @@ components:
...
+        a way to set a default download level at registration that is copied
+        down to all the files.
+
+      type: string
+      enum:
+        - 'world'
+        - 'stanford'
+        - 'location-based'
+        - 'none'
+      default: 'none'
```
justinlittman, line 128: Is `none` correct or should this be nullable?

openapi.yml
```diff
@@ -99,36 +99,54 @@ components:
...
+        - world
+        - stanford
+        - location-based
+        - citation-only
+        - dark
+      default: 'dark'
+    AccessReadLocation:
+      description: If access is "location-based", which location should have access.
+      type: string
+      nullable: true
```
justinlittman, line 142: Why is this nullable?

openapi.yml
```diff
@@ -99,36 +99,54 @@ components:
...
+    AccessLevel:
+      description: Access level that applies when embargo expires.
+      type: string
+      enum:
+        - world
+        - stanford
+        - location-based
+        - citation-only
+        - dark
+      default: 'dark'
```
justinlittman, line 138: This didn't have a default everywhere. Can it?

openapi.yml
justinlittman: This didn't have a default everywhere. Can it?

lib/cocina/models/access_read_location.rb
```diff
@@ -0,0 +1,9 @@
+# frozen_string_literal: true
+
+module Cocina
+  module Models
+    AccessReadLocation = Types::String.constrained(
```
jcoyne, line 5: This constant is never actually used in the ruby, right?

jcoyne (approved):

openapi.yml
jcoyne: sure.

openapi.yml
jcoyne: Because you might want to clear it out if it was previously set to a value.

openapi.yml
jcoyne: sure.

lib/cocina/models/access.rb
```diff
@@ -3,13 +3,13 @@
 module Cocina
   module Models
     class Access < Struct
-      # Access level
+      # Access level that applies when embargo expires.
```
jcoyne, line 6: ⚪️ This comment doesn't seem correct. This is the current access level, right?

lib/cocina/models/dro_access.rb
```diff
@@ -3,6 +3,7 @@
 module Cocina
   module Models
     class DROAccess < Struct
+      # Access level that applies when embargo expires.
```
jcoyne, line 6: ⚪️ this comment doesn't seem right either.

lib/cocina/models/file_access.rb
```diff
@@ -3,14 +3,15 @@
 module Cocina
   module Models
     class FileAccess < Struct
-      # Access level
+      # Access level that applies when embargo expires.
```
jcoyne, line 6: ⚪️ and here

jcoyne (changes requested):

PR #255, opened 2021-04-16
More precise modeling of DRO and embargo rights. Adds generator suppo…

### LINKED ISSUE #253: embargo release access rights values should support all values in the Argo rights menu

justinlittman:
…rt for oneOf.

closes #253

## Why was this change made?
* Improved validation of access rights for DRO and embargoes.

Unit

openapi


justinlittman:
Like other cocina model PRs, the test coverage is wonky.

PR #257, opened 2021-04-16
[HOLD] More precise modeling of rights

justinlittman:

## Why was this change made?
* Improved validation of access rights for DRO, embargoes, and files.

Unit

openapi


lib/cocina/generator/schema_base.rb
```diff
@@ -54,6 +55,12 @@ def example
         "# example: #{schema_doc.example}\n"
       end
 
+      def relaxed_comment
+        return '' unless relaxed
+
+        "# Validation of this property is relaxed. See the openapi for full validation.\n"
```
jcoyne, line 61: 👍

jcoyne (approved):
This seems like a good change. But I'm a little hung up on the "Validation of this property is relaxed" statement.  I'm not sure that communicates what is actually happening. Does anybody have thoughts on how that could be made clearer?  How about "This property may be required depending on how other values in this struct are set?"

ndushay:
so this is merged, but it still says "HOLD".   I'm also unclear as to why you didn't test with various code bases pointing to the branch for this PR ... before it was merged.

jcoyne:
I don't think it's possible to point all the codebases at a prerelease of cocina models, because of how there are transitive dependencies with dor-services-client and sdr-client gems (which cannot express a dependency on a git release).  We've always cut a new version, then evaluated it, and then decided whether or not to deploy it.

PR #260, opened 2021-05-07
Move hasAgreement to AdminPolicy

### LINKED ISSUE #259: DROStructural:  no agreements, only APOs have them.  APOs should REQUIRE agreementId
1.  Currently DROStructural contains "hasAgreement" but per @andrewjbtw:

> items don’t need to have agreements anywhere
> the reference back to agreements is just supposed to be in the APO RELS-EXT

2. require agreementId for AdminPolicy objects (in adminPolicyAdminstrative as "referencesAgreement" )

jcoyne:

## Why was this change made?
Fixes #259

### DIFF
--- a/docs/maps/Collection.json
+++ b/docs/maps/Collection.json
@@ -215,12 +215,7 @@
     "structural": {
       "description": "Structural metadata for the Collection.",
       "type": "object",
-      "required": ["hasAgreement"],
       "properties": {
-        "hasAgreement": {
-          "description": "Agreement that covers the deposit of the Collection into SDR.",
-          "type": "string"
-        },
         "hasMember": {
           "description": "Component digital repository objects or collections that are a part of this collection.",
           "type": "array",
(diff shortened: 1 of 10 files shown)


justinlittman (changes requested):
So this requires `hasAgreement`. However, the ticket suggests Honeybadger alerting when an APO is missing an agreement, suggesting that some APOS do not have agreements. Is that going to be a problem?

justinlittman:
P.S. That is my only issue with this PR. Otherwise, I approve.

jcoyne:
@justinlittman item 2 and item 3 on that issue seem to be contradictory.  What do you think we should do?

justinlittman:
We could write a report to check if there are APOs missing an agreement and then decide to remediate or lax enforce.

justinlittman:
https://github.com/sul-dlss/dor-services-app/issues/3142 is for the report.

justinlittman (approved):

PR #261, opened 2021-05-10
Allow copyright statement to be nullable

jcoyne:

## Why was this change made?
So that it can be cleared out if it is not desired


jcoyne:
Part of the fix is in https://github.com/sul-dlss/cocina-models/pull/261

justinlittman (approved):

PR #263, opened 2021-06-04
Enumerate the valid license values

jcoyne:
## Why was this change made?

This should help prevent data errors like https://github.com/sul-dlss/happy-heron/issues/1584


justinlittman (changes requested):
Regenerate models?

jcoyne:
@justinlittman thanks, I've added that.

justinlittman (approved):

PR #266, opened 2021-06-24
Validate that PURL starts with http, not https.

### LINKED ISSUE #265: Validate that PURL starts with http, not https.
Canonical PURLs start with http. Non-canonical PURLs cause problems for roundtrip validation.

justinlittman:
closes #265

## Why was this change made?
Enforce canonical PURLs.

Unit

openapi.yml

### DIFF
--- a/openapi.yml
+++ b/openapi.yml
@@ -582,9 +582,7 @@ components:
           items:
             $ref: "#/components/schemas/DescriptiveValue"
         purl:
-          description: Stanford persistent URL associated with the resource.
-          type: string
-          format: uri
+          $ref: "#/components/schemas/Purl"
         access:
           $ref: "#/components/schemas/DescriptiveAccessMetadata"
         relatedResource:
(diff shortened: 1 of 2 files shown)


openapi.yml
```diff
@@ -1389,6 +1387,11 @@ components:
         width:
           description: Width in pixels
           type: integer
+    Purl:
+      description: Stanford persistent URL associated with the related resource.
+      type: string
+      format: uri
+      pattern: '^http://'
```
jcoyne, line 1394: Would you mind dropping in a comment here that says "Arcadia has stated that the canonical URL will be http, even though this redirects to https"

jcoyne (changes requested):

openapi.yml
justinlittman: Added.

jcoyne (approved):

PR #270, opened 2021-07-06
Add a field for DOI in the Identification schema

### LINKED ISSUE #269: Add DOI to identification
See https://github.com/sul-dlss/cocina-models/blob/257f92d8fdf4cf5a77359dca372f43ced6fbc53c/docs/maps/DRO.json#L238

jcoyne:
## Why was this change made?

Fixes #269 
Needed so that h2 can transfer the object and it's associated DOI to DSA.

### DIFF
--- a/lib/cocina/models/purl.rb
+++ b/lib/cocina/models/purl.rb
@@ -1,0 +1,9 @@
+# frozen_string_literal: true
+
+module Cocina
+  module Models
+    Purl = Types::String.constrained(
+      format: %r{^http://}i
+    )
+  end
+end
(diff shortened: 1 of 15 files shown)


lib/cocina/models/purl.rb
```diff
@@ -0,0 +1,9 @@
+# frozen_string_literal: true
+
+module Cocina
+  module Models
+    Purl = Types::String.constrained(
```
jcoyne, line 5: This file didn't get generated in #266

justinlittman (approved):

PR #272, opened 2021-07-07
Add DOI to the zeitwerk inflector

jcoyne:

## Why was this change made?

otherwise, when an application eager loads (in production mode), we get this error:
```
Could not spawn process for application /opt/app/sdr/sdr-api/current: The application encountered the following error: expected file /opt/app/sdr/sdr-api/shared/bundle/ruby/2.6.0/gems/cocina-models-0.61.0/lib/cocina/models/doi.rb to define constant Cocina::Models::Doi, but didn’t (Zeitwerk::NameError)
```

### DIFF
--- a/lib/cocina/models.rb
+++ b/lib/cocina/models.rb
@@ -14,27 +14,19 @@ require 'thor'
 
 # Help Zeitwerk find some of our classes
 class CocinaModelsInflector < Zeitwerk::Inflector
-  # rubocop:disable Metrics/MethodLength
-  def camelize(basename, _abspath)
-    case basename
-    when 'dro'
-      'DRO'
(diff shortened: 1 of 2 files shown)


lib/cocina/models.rb
```diff
@@ -14,27 +14,19 @@
...
-    when 'dro_structural'
-      'DROStructural'
-    when 'request_dro_structural'
-      'RequestDROStructural'
-    when 'version'
-      'VERSION'
-    else
-      super
-    end
+    INFLECTIONS.fetch(basename) { super }
```
aaron-collier, line 28: :+1: to this!

aaron-collier (approved):

PR #273, opened 2021-07-08
Adapt the DOI pattern to omit the druid namespace

jcoyne:

## Why was this change made?
This follows our existing dois: 10.25740/xd282zn6829 for example


ndushay (approved):
yup - makes sense.

PR #282, opened 2022-01-06
Switched to inline orb for circleci.

justinlittman:

## Why was this change made?
Clean, lean


edsu:
This can be closed since #283 was merged.

PR #283, opened 2022-01-12
Switch to sul-dlss orb for circleci config.

edsu:

## Why was this change made?

Reduce configuration.

### DIFF
--- a/Gemfile
+++ b/Gemfile
@@ -5,7 +5,5 @@ source 'https://rubygems.org'
 # Specify your gem's dependencies in cocina-models.gemspec
 gemspec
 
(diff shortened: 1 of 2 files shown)


Gemfile
```diff
@@ -5,4 +5,7 @@ source 'https://rubygems.org'
 # Specify your gem's dependencies in cocina-models.gemspec
 gemspec
 
-gem 'byebug'
+group :development, :test do
```
jcoyne, line 8: A development and test group doesn't make a ton of sense in the context of a gem since we never deploy this.

justinlittman (approved):

PR #290, opened 2022-01-19
Changes array to normalize to empty array.

justinlittman:

## Why was this change made?
So that logically equivalent Cocina models normalize to equivalent hashes.

Unit


spec/cocina/models/administrative_spec.rb
```diff
@@ -0,0 +1,26 @@
...
+    end
+
+    let(:administrative2) do
+      described_class.new({
+                            hasAdminPolicy: 'druid:bc123df4567',
+                            releaseTags: []
+                          })
+    end
+
+    it 'normalizes array' do
```
justinlittman, line 20: This shows why the change is desirable.

jcoyne (approved):

PR #295, opened 2022-01-25
Requires PURL for non-request descriptions.

### LINKED ISSUE #293: Require PURL
Require PURL for Description for DROs, Collections, and AdminPolicies (but not Request*).

See #291.

justinlittman:
closes #293

## Why was this change made?
Data validation.

unit

openapi

### DIFF
--- a/openapi.yml
+++ b/openapi.yml
@@ -91,6 +91,18 @@ paths:
       responses:
         '200':
           description: noop
+  /validate/RequestDescription:
+    post:
+      summary: Validate a Request Description
+      requestBody:
+        required: true
+        content:
+          application/json:
+            schema:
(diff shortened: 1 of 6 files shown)


openapi.yml
```diff
@@ -1556,6 +1506,78 @@ components:
         - label
         - type
         - version
+    RequestDescription:
```
ndushay, line 1509: this could have a description itself, like "same as Description but without purl property" or "Description to be sent as part of a Request object" or something.

ndushay (changes requested):
requesting you add a `description` in `openapi.yml` for `RequestDescription`

ndushay (approved):

PR #303, opened 2022-02-04
[SPIKE] Create an interface for effectively mutating model instances

mjgiarlo:
Working with our cocina-models can be jarring because, unlike most of the rest of the code we write, dry-structs are immutable which can lead to surprising results when leaning on the usual assignment interactions. So, try something different. What is in this commit (which is partly copypasta from a dry-struct issue) works fine in the simplest use cases (see administrative spec), but currently fails in the gnarlier use cases (file access spec) particularly when reaching into a structure (like a DRO) to modify another structure (like a File). Why? Because File instances are not self-validating. Files are validated in the context of the DRO, and when you reach into the DRO and mutate the File, I have so far not found a way to force a validation in the surrounding context. This feels a lot like working against dry-struct/-types, too, so I am ending the spike and pushing this up for comments.

### DIFF
--- a/lib/cocina/models.rb
+++ b/lib/cocina/models.rb
@@ -49,6 +49,18 @@ module Cocina
     class Struct < Dry::Struct
       transform_keys(&:to_sym)
       schema schema.strict
+
+      def self.attribute(name, type = nil, &block)
+        super
+
+        define_method("#{name}=") do |value|
+          self.attributes = attributes.merge(name => value)
+        end
+      end
+
+      def attributes=(new_attributes)
+        @attributes = self.class.schema[new_attributes]
+      end
     end
 
     # DRY Types
--- a/spec/cocina/models/administrative_spec.rb
+++ b/spec/cocina/models/administrative_spec.rb
@@ -23,4 +23,19 @@ RSpec.describe Cocina::Models::Administrative do
       expect(administrative1.to_h).to eq(administrative2.to_h)
     end
   end
+
+  describe 'mutation' do
+    let(:administrative) do
+      described_class.new(
+        {
+          hasAdminPolicy: 'druid:bc123df4567'
+        }
+      )
+    end
+
+    it 'provides a setter as a mutation facade' do
+      administrative.hasAdminPolicy = 'druid:cb321fd7654'
+      expect(administrative.hasAdminPolicy).to eq('druid:cb321fd7654')
+    end
+  end
 end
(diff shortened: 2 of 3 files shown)


spec/cocina/models/administrative_spec.rb
```diff
@@ -23,4 +23,19 @@
...
+    let(:administrative) do
+      described_class.new(
+        {
+          hasAdminPolicy: 'druid:bc123df4567'
+        }
+      )
+    end
+
+    it 'provides a setter as a mutation facade' do
+      administrative.hasAdminPolicy = 'druid:cb321fd7654'
```
mjgiarlo, line 37: Easy-peasy. Works like a charm.

spec/cocina/models/file_access_spec.rb
```diff
@@ -4,49 +4,51 @@
...
+                                                       access: {},
+                                                       hasMessageDigests: []
+                                                     }
+                                                   ]
+                                                 }
+                                               }
+                                             ]
+                                           })
+
+    dro_template.structural.contains.first.structural.contains.first.access = {
```
mjgiarlo, line 45: The operation succeeds but the DRO fails to raise a validation error because a child structure was mutated and the child has no programmatic connection up to the parent to force a validation in the containing context, at least none that I can find.

lib/cocina/models.rb
```diff
@@ -49,6 +49,18 @@ class ValidationError < Error; end
...
+        super
+
+        define_method("#{name}=") do |value|
+          self.attributes = attributes.merge(name => value)
+        end
+      end
+
+      def attributes=(new_attributes)
+        @attributes = self.class.schema[new_attributes]
+      end
```
mjgiarlo, line 63: Mostly copypasta with an update or two.

spec/cocina/models/file_access_spec.rb
mjgiarlo: This may be related to file access validation being "relaxed," *i.e.* done by OpenAPI instead of by dry-struct? 

Worth a deeper look at a later date, perhaps. But perhaps not given the mutator interface reaches into the guts of dry-struct with relatively little payoff.

Cc: @justinlittman

justinlittman: Both of these are true. Only some cocina models are validated on instantiation. And only some validation is performed by dry-struct (the OpenAPI validation is more expressive / powerful).

justinlittman:
@mjgiarlo Can this be closed?

mjgiarlo:
@justinlittman Yep!

PR #309, opened 2022-02-11
Model virtual object relationship on the constituent side

mjgiarlo:

## Why was this change made?

This lets us generate public XML relating constituents back to their virtual objects.

CI

None


mjgiarlo:
Don't need this! We'll indexing the relationship in Solr and query it from there.

PR #314, opened 2022-02-25
Validates on instantiation from existing instance.

### LINKED ISSUE #313: Add enums to the structs when enums are in the openapi.yml
See https://dry-rb.org/gems/dry-types/1.0/enum/

Start with file access.  Once complete, this should raise an error:

```ruby
Cocina::Models::FileAccess.new(access: 'citation-only', download: 'none')
```

justinlittman:
closes #313

Catch validation problems faster.

Unit

### DIFF
--- a/lib/cocina/generator/schema.rb
+++ b/lib/cocina/generator/schema.rb
@@ -16,11 +16,10 @@ module Cocina
             module Models
               class #{name} < Struct
 
+                #{validate}
                 #{types}
 
                 #{model_attributes}
-
-                #{validate}
               end
             end
           end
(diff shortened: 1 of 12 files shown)


lib/cocina/generator/schema.rb
```diff
@@ -67,7 +67,14 @@ def validate
           def self.new(attributes = default_attributes, safe = false, validate = true, &block)
             Validator.validate(self, attributes.with_indifferent_access) if validate && name
             super(attributes, safe, &block)
-                end
+          end
+
+          def new(*args)
```
jcoyne, line 72: Can we put this code in a module that we include on the model classes?

jcoyne (changes requested):

jcoyne (approved):

PR #337, opened 2022-03-04
Update URIs to be https and drop jsonld

### LINKED ISSUE #329: Change structure of vocab URLs
Make `https` and remove `.jsonld`. 

For example:
```
https://cocina.sul.stanford.edu/models/admin_policy
```

jcoyne:

Fixes #329


lib/cocina/models/collection.rb
```diff
@@ -7,11 +7,11 @@ class Collection < Struct
...
-      TYPES = ['http://cocina.sul.stanford.edu/models/collection.jsonld',
-               'http://cocina.sul.stanford.edu/models/curated-collection.jsonld',
-               'http://cocina.sul.stanford.edu/models/user-collection.jsonld',
-               'http://cocina.sul.stanford.edu/models/exhibit.jsonld',
-               'http://cocina.sul.stanford.edu/models/series.jsonld'].freeze
+      TYPES = ['https://cocina.sul.stanford.edu/models/collection',
+               'https://cocina.sul.stanford.edu/models/curated-collection',
+               'https://cocina.sul.stanford.edu/models/user-collection',
+               'https://cocina.sul.stanford.edu/models/exhibit',
+               'https://cocina.sul.stanford.edu/models/series'].freeze
```
mjgiarlo, line 14: If we remove the old URIs from these lists right now, won't lots of objects be confused as to what they are?

mjgiarlo (commented):
I have a question about existing type URIs being removed, specifically about the `Checkable` module's downstream consumers: DSA in particular asks Cocina objects whether they are DROs or Collections (etc.) and if the old-style type URIs are persisted but the new-style ones are in the gem, DSA is going to be confused. (Right?)

lib/cocina/models/collection.rb
jcoyne: No, because we don't persist these in Fedora yet. That's why it's important to make these changes now.

mjgiarlo (approved):
🚀

Issue #311, opened 2022-02-18
Add TitleBuilder and RightsDescriptionBuilders as utility methods


jcoyne:
This would include extracting TitleBuilder: https://github.com/sul-dlss/dor-services-app/blob/main/app/services/title_builder.rb and *RightsDescriptionBuilders (e.g. https://github.com/sul-dlss/dor_indexing_app/blob/main/app/services/dro_rights_description_builder.rb) We also might consider putting the MODS<-> cocina mappings in there. 

This is helpful because we already have copied code between DSA and DIA for building titles.  If we have this as an available library it could allow easier adoption of the cocina model by the access team and would allow for refactoring in Argo.

mjgiarlo:
Additionally, I wonder if we want cocina-models to define one or more RSpec matchers for downstream codebases. We are already seeing these duplicated (DSA, Argo).

ndushay:
I like Mike's idea about rspec matches in cocina-models as needed.   Can that be ticketed (for specific matchers we already use?)

ndushay:
In fact, can all the re-used "interpret this data in the cocina model" code be ticketed ?

jcoyne:
@mjgiarlo feel free to make matchers as a separate ticket.

mjgiarlo:
@jcoyne 💬 
> @mjgiarlo feel free to make matchers as a separate ticket.

https://github.com/sul-dlss/cocina-models/issues/319

PR #342, opened 2022-03-07
Extract DSA TitleBuilder into Cocina::Services::TitleBuilder

aaron-collier:

Part of #311 

Added Unit Tests

### DIFF
--- a/lib/cocina/models/admin_policy.rb
+++ b/lib/cocina/models/admin_policy.rb
@@ -7,6 +7,8 @@ module Cocina
 
       include Checkable
 
+      include TitleBuilder
+
       TYPES = ['https://cocina.sul.stanford.edu/models/admin_policy'].freeze
 
       # The version of Cocina with which this object conforms.
--- a/lib/cocina/models/admin_policy.rb
+++ b/lib/cocina/models/admin_policy.rb
@@ -7,8 +7,6 @@ module Cocina
 
       include Checkable
 
-      include TitleBuilder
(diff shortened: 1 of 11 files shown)


lib/cocina/models/admin_policy.rb
```diff
@@ -7,6 +7,8 @@ class AdminPolicy < Struct
 
       include Checkable
 
+      include TitleBuilder
```
jcoyne, line 10: I don't think we should include these behaviors on the models.  Can these just be services?

jcoyne (changes requested):

lib/cocina/models/admin_policy.rb
aaron-collier: Yeah - totally, I was actually thinking that would be a good idea but would like a short pairing to discuss.

aaron-collier:
@jcoyne marking this as draft while I work on improving the test coverage and address a bug or two that has come up in doing so.

lib/cocina/services/title_builder.rb
```diff
@@ -0,0 +1,203 @@
+# frozen_string_literal: true
+
+module Cocina
+  module Services
```
jcoyne, line 4: I think this should still be `models` since this is the `cocina-models` gem.

jcoyne (changes requested):

jcoyne (approved):

PR #353, opened 2022-03-11
Add RightDescriptionBuilder, DroRightsDescriptionBuilder, and test coverage for both

### LINKED ISSUE #311: Add TitleBuilder and RightsDescriptionBuilders as utility methods

aaron-collier:

Closes #311 

Unit tests added.

### DIFF
--- a/lib/cocina/models/dro_rights_description_builder.rb
+++ b/lib/cocina/models/dro_rights_description_builder.rb
@@ -1,0 +1,67 @@
+# frozen_string_literal: true
+
+module Cocina
+  module Models
+    # Rights description builder for items
+    class DroRightsDescriptionBuilder < RightsDescriptionBuilder
+      # @param [Cocina::Models::DRO] cocina_item
+
+      # This overrides the superclass
+      # @return [Cocina::Models::DROAccess]
+      def object_access
+        @object_access ||= cocina.access
+      end
+
+      private
+
+      def object_level_access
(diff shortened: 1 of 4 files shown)


lib/cocina/models/dro_rights_description_builder.rb
```diff
@@ -0,0 +1,67 @@
+# frozen_string_literal: true
+
+module Cocina
+  module Models
+    # Rights description builder for items and apos
```
jcoyne, line 5: Should apos be removed here?

jcoyne (changes requested):

lib/cocina/models/dro_rights_description_builder.rb
aaron-collier: Cruft. Fixed, thanks for catching that.

jcoyne (approved):

PR #354, opened 2022-03-14
Add RSpec matchers for cocina comparisons

### LINKED ISSUE #319: Define Cocina RSpec matchers
Occasionally when downstream codebases are testing Cocina structures, they are creating custom Cocina matchers. These matchers could live in the cocina-models gem so that they don't have to live everywhere else. Ensure each of the above codebases' test suites pass once they pick this up from cocina-models:

- [x] https://github.com/sul-dlss/dor-services-app/blob/main/spec/support/cocina_matchers.rb
- [x] https://github.com/sul-dlss/argo/blob/main/spec/support/cocina_matchers.rb
- [x] https://github.com/sul-dlss/sdr-client/blob/main/spec/support/cocina_matchers.rb

Also:

- [x] Describe the matcher(s) in the README

mjgiarlo:
Fixes #319

This commit ports Cocina matchers from other codebases (argo, dor-services-app, sdr-client) so that they are available downstream. Companion PRs in these codebases are in progress and will be ready for review once we cut the next release of cocina-models.

CI, and this change was already tested successfully in argo, dor-services-app, and sdr-client.

### DIFF
--- a/.rubocop.yml
+++ b/.rubocop.yml
@@ -78,6 +78,7 @@ Metrics/BlockLength:
   Exclude:
     - cocina-models.gemspec
     - spec/cocina/**/*
+    - lib/cocina/rspec/matchers.rb
 
 Metrics/MethodLength:
   Max: 14
--- a/README.md
+++ b/README.md
@@ -103,3 +103,17 @@ The following are the recommended naming conventions for code using Cocina model
 * `cocina_admin_policy`: `Cocina::Models::AdminPolicy` instance
 * `cocina_collection`: `Cocina::Models::Collection` instance
 * `cocina_object`: `Cocina::Models::DRO` or `Cocina::Models::AdminPolicy` or `Cocina::Models::Collection` instance
+
+## RSpec matchers
+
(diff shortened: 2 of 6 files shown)


edsu:
Would it be useful to pull some of the ones in https://github.com/sul-dlss/argo/blob/main/spec/support/cocina_matchers.rb too?

mjgiarlo:
@edsu 💬 
> Would it be useful to pull some of the ones in [https://github.com/sul-dlss/argo/blob/main/spec/support/cocina_matchers.rb](https://github.com/sul-dlss/argo/blob/main/spec/support/cocina_matchers.rb?rgh-link-date=2022-03-14T17%3A30%3A32Z) too?

Yep! I think I got 'em all. What's not obvious is that `cocina_object_with` handles all of what `cocina_object_with_access`, `cocina_object_with_administrative`, and `cocina_object_with_identification` handled.

edsu (approved):
I like the idea of consolidating these here for use elsewhere in other repos. Also, thanks for the examples in the README.

PR #359, opened 2022-03-15
Update README with new CI-based cocina level 2 PR creation notes

mjgiarlo:


README.md
```diff
@@ -53,10 +53,14 @@ which pushes the gem to rubygems.org.  Next write up the release notes: https://
 
 ### Step 2: Update client gems coupled to the models
 
+**NOTE**: You can skip this step if the new release is a patch-level bump only.
```
justinlittman, line 56: Are you sure? https://github.com/sul-dlss/sdr-api/blob/main/app/controllers/resources_controller.rb#L73-L81

justinlittman (changes requested):

README.md
jcoyne: @justinlittman the client libraries don't need to be updated, but the consumers need to have `bundle update cocina-models`, because the pin in the clients float at the patch level, there's no need to update them.

mjgiarlo: What @jcoyne said. And that is done in steps 3b and 4.

justinlittman (approved):

PR #361, opened 2022-03-18
Optimistic locking

justinlittman:

This provides a mechanism for holding lock, create date, and modified date in a manner that is compatible with existing DRO, AdminPolicy, and Collection classes.

Unit, integration tests

### DIFF
--- a/lib/cocina/generator/schema.rb
+++ b/lib/cocina/generator/schema.rb
@@ -47,10 +47,13 @@ module Cocina
       end
 
       def types
-        type_properties_doc = schema_doc.properties['type']
-        return '' if type_properties_doc.nil? || type_properties_doc.enum.nil?
+        type_schema_property = schema_properties.find { |schema_property| schema_property.key == 'type' }
+        return '' if type_schema_property.nil?
 
-        types_list = type_properties_doc.enum.map { |item| "'#{item}'" }.join(",\n ")
+        type_schema_doc = type_schema_property.schema_doc
+        return '' if type_schema_doc.enum.nil?
+
+        types_list = type_schema_doc.enum.map { |item| "'#{item}'" }.join(",\n ")
 
         <<~RUBY
           include Checkable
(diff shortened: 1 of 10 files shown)


lib/cocina/generator/schema.rb
```diff
@@ -47,10 +47,13 @@ def model_attributes
       end
 
       def types
-        type_properties_doc = schema_doc.properties['type']
-        return '' if type_properties_doc.nil? || type_properties_doc.enum.nil?
+        type_schema_property = schema_properties.find { |schema_property| schema_property.key == 'type' }
```
justinlittman, line 50: This is necessary because `*WithMetadata` use `allOf`. This case wasn't previously handled for generating types.

lib/cocina/rspec/matchers.rb
```diff
@@ -11,16 +11,21 @@ module Matchers
...
           match do |actual|
-            expected.all? do |expected_key, expected_value|
-              # NOTE: there's no better method on Hash that I could find for this.
-              #        #include? and #member? only check keys, not k/v pairs
-              actual.public_send(cocina_section).to_h.any? do |actual_key, actual_value|
-                if expected_value.is_a?(Hash) && actual_value.is_a?(Hash)
-                  expected_value.all? { |pair| actual_value.to_a.include?(pair) }
-                else
-                  actual_key == expected_key && actual_value == expected_value
+            # created, modified, lock for *WithMetadata don't respond to all?
```
justinlittman, line 14: L14 explains this change.

spec/spec_helper.rb
```diff
@@ -2,6 +2,7 @@
 
 require 'bundler/setup'
 require 'cocina/models'
+require 'cocina/rspec'
```
justinlittman, line 5: This allows using the matchers in this app, which allows testing the above change to the matcher.

jcoyne (approved):

PR #366, opened 2022-03-22
Title builder should handle an array of Titles or DescriptiveValue

jcoyne:
This is required in dor-indexing-app for https://github.com/sul-dlss/dor_indexing_app/issues/796
Because it is used by the TopicBuilder here: https://github.com/sul-dlss/dor_indexing_app/blob/main/app/services/topic_builder.rb#L78

test suite


justinlittman (approved):
Uggh.

PR #368, opened 2022-03-23
README: add a naming convention for DROs that are agreements.

ndushay:

agreement objects are a thing

README changes only.


mjgiarlo (approved):
LGTM

PR #372, opened 2022-03-30
Make sourceId required for DROs

### LINKED ISSUE #369: Make identification.sourceId be a required property for DROs
https://github.com/sul-dlss/cocina-models/blob/main/openapi.yml#L1245-L1258

jcoyne:

Fixes #369

unit tests


jcoyne:
@justinlittman I'm just going on what Andrew said.

andrewjbtw:
Not sure about any environment but unless something is hiding from the indexer in prod, I went through and assigned source ids to every item without them last year. They should show up as "Cocina conversion failed" in the Data Quality facet if a source id is missing.

PR #373, opened 2022-03-30
Add a property to indicate the catalog link is used for refresh

### LINKED ISSUE #370: Add a refresh-from-catkey field
Currently we allow multiple catkeys, but we have behavior in code where it only uses the first catkey when doing a refresh of the metadata.  This behavior is opaque in the data.  Instead we should have a special catkey for refreshing that we only allow one of.

This comes from a discussion on slack between Arcadia, Andrew and Justin C on 2022-03-29

jcoyne:

Fixes #370


openapi.yml
```diff
@@ -369,18 +369,28 @@ components:
...
       properties:
         catalog:
           description: Catalog that is the source of the linked record.
           type: string
+          enum:
+            - symphony
+            - previous symphony
           example: symphony
+        refresh:
+          description: Only one of the catkeys should be designated for refreshing.
```
justinlittman, line 381: Can this be modeled in a manner in which this is validatable?

justinlittman (changes requested):

openapi.yml
justinlittman: I'd further suggest that whether the descriptive metadata can be refreshed should be an attribute of description, not identification. If my understanding is correct it is saying "this description has been modified so don't refresh it", not "this catkey should not be used to perform a refresh".

jcoyne: @justinlittman I don't know how to model that it may have 0..1 that have refresh:true.  Do you have any idea?  I believe moving catalogLinks out of identification is beyond the scope of this ticket.

justinlittman: I wasn't suggesting that catalogLinks should be moved out of identification. My suggestion would be:
1. Leaving catalogLinks the way it is and add documentation that states that the first "symphony" link is the link that is used for refresh.
2. Add a boolean "allow_refresh" (or similar) attribute to description.

I'd further suggest that we just do 1 now and treat 2 as a completely separate enhancement (as it is new functionality and will require changes in other codebases).

justinlittman (approved):

PR #374, opened 2022-03-30
Generate snake_case names in Rubyland (with backwards compat via generation + method_missing)

### LINKED ISSUE #335: Support snakecase
instead of current camelcase.

Use of camelcase should be allowed, but emit a deprecation warning.

Set deprecation horizon to 1.0 or 2.0 or something

mjgiarlo:

This commit allows models to keep JavaScript naming conventions in JSland (using camelCase for JSON-based Cocina data structures, e.g., in OpenAPI specifications) and Ruby naming conventions in Rubyland (using snake_case for Ruby bindings, e.g., hashes and method calls).

Fixes #335

Includes:

* Convert attributes from camelCase to snake_case at generate time
* Coerce attributes to camelCase when validating against the OpenAPI specification (i.e., using JSland conventions)
* Generate `self.new` method into models even if they are not validatable to allow us to coerce attributes to underscore form (for Rubyland joy) at runtime
* Add `#method_missing` to all affected models (via generation) to allow downstream codebases to continue using camelCase in Rubyland (backwards compatibility)
  * And leave tests untouched to demonstrate backwards compatibility
* Deprecate and warn that backwards compatibility goes away as of version 1.0.0
* Regenerate all models
* Regenerate Rubocop TODOs

CI


mjgiarlo:
Superseded by https://github.com/sul-dlss/cocina-models/pull/375 (don't bother with metaprogramming, not needed).

PR #375, opened 2022-03-30
[HOLD] Generate snake_case names in Rubyland

### LINKED ISSUE #335: Support snakecase

mjgiarlo:
This is on hold until the week of April 18th, after other Cocina changes have gone out.

This commit allows models to keep JavaScript naming conventions in JSland (using camelCase for JSON-based Cocina data structures, e.g., in OpenAPI specifications) and Ruby naming conventions in Rubyland (using snake_case for Ruby bindings, e.g., hashes and method calls).

Fixes #335

Includes:

* Convert attributes from camelCase to snake_case at generate time
* Coerce attributes to camelCase when validating against the OpenAPI specification (i.e., using JSland conventions)
* Generate `self.new` method into models even if they are not validatable to allow us to coerce attributes to underscore form (for Rubyland joy) at runtime
* Add aliases (with deprecations) to all affected models to allow downstream codebases to continue using camelCase in Rubyland (backwards compatibility)
  * And leave tests untouched to demonstrate backwards compatibility and deprecation warnings
* Deprecate and warn that backwards compatibility goes away as of version 1.0.0
* Regenerate all models
* Regenerate Rubocop TODOs

CI


jcoyne (approved):

mjgiarlo:
We decided not to move forward with camels-to-snakes translation because the translation, we learned, leaks out of cocina-models to at least DSC and DSA and possibly many other locations. And given that concern and the timing of our work to replace Fedora and persist Cocina JSON directly, we are ending this work.

PR #378, opened 2022-03-31
Refactors validators and adds dark validator.

### LINKED ISSUE #377: Move validate dark service to cocina models
So that it is performed on every instantiation rather than just update/create.

justinlittman:
closes #377

To move semantic validation as close to data creation as possible.

Unit


lib/cocina/generator/generator.rb
```diff
@@ -78,7 +78,6 @@ def generate_for(schema)
         'rights_description_builder.rb',
         'title_builder.rb',
         'validatable.rb',
-        'validator.rb',
```
mjgiarlo, line 81: Presumably we don't need to add the new validators to this list because they're in their own subdirectory?

justinlittman: As it turns out, yes.

lib/cocina/models/validatable.rb
```diff
@@ -8,15 +8,15 @@ module Validatable
...
+          Validators::Validator.validate(self, attributes.with_indifferent_access) if validate
           super(attributes, safe, &block)
         end
       end
 
       def new(*args)
         validate = args.first.delete(:validate) if args.present?
         new_model = super(*args)
-        Validator.validate(new_model.class, new_model.to_h) if (validate || validate.nil?) && self.class.name
+        Validators::Validator.validate(new_model.class, new_model.to_h) if validate || validate.nil?
```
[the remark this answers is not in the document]
justinlittman: I was trying to figure out what it did and realized that it did nothing.

mjgiarlo (approved):

lib/cocina/models/validators/dark_validator.rb
```diff
@@ -0,0 +1,79 @@
...
+
+        def meets_preconditions?
+          dro? && attributes.dig(:access, :view) == 'dark'
+        end
+
+        def dro?
+          (clazz::TYPES & DRO::TYPES).any?
+        rescue NameError
+          false
+        end
```
mjgiarlo, line 38: @justinlittman I'm wondering if we can call `clazz.dro?` and lean on the `Checkable` module for this instead.

mjgiarlo: Aha, it's because it's an instance method and we have but a class and some props. Nevermind!

PR #380, opened 2022-04-01
Description type validation

justinlittman:

To have valid data you have to validate it.

Unit

### DIFF
--- a/description_types.yml
+++ b/description_types.yml
@@ -1,0 +1,550 @@
+# Description type validation
+
+access.accessContact:
+  - value: email
+    description: Email address for a contact person or institution concerning the resource.
+  - value: repository
+    description: Institution providing access to the resource.
+access.digitalLocation:
+  - value: discovery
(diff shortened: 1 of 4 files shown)


justinlittman:
@arcadiafalcone I'm going to leave this as a draft until you've had a chance to make changes to `description_types.yml`. Feel free to make PRs against this branch.

description_identifier_types.yml
```diff
@@ -0,0 +1,10 @@
+# Description identifier type validation
+
+identifier: # expand list based on mapping
```
jcoyne, line 3: Can we create a DescriptiveIdentifier class that is like a DescriptiveValue, but only supports these values as an enum?  Then we can just use the openapi validation and we don't have to write a custom validator.

https://github.com/sul-dlss/cocina-models/blob/main/openapi.yml#L1685-L1689

justinlittman:
Before:

jcoyne (approved):

PR #389, opened 2022-04-08
Add validator ensuring no more than one refresh catalog link is present

### LINKED ISSUE #379: CatalogLinks validator
As follow-up to https://github.com/sul-dlss/cocina-models/pull/373, write a validator that validates that there is only a catalog link with refresh set to true.

mjgiarlo:
Fixes #379

To continue adding smarter semantic validation to cocina-models, and this is the next step. Having a single catalog link with the refresh property set to true makes it clear which catalog link ("catkey") to use for refreshing metadata.

CI

### DIFF
--- a/lib/cocina/models/validators/catalog_links_validator.rb
+++ b/lib/cocina/models/validators/catalog_links_validator.rb
@@ -1,0 +1,54 @@
+# frozen_string_literal: true
+
+module Cocina
+  module Models
+    module Validators
+      # Validates that only a single CatalogLink has refresh set to true
+      class CatalogLinksValidator
+        MAX_REFRESH_CATALOG_LINKS = 1
+
+        def self.validate(clazz, attributes)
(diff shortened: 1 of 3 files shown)


justinlittman (approved):

arcadiafalcone:
@mjgiarlo @justinlittman One exception is objects with the tag "Warning : Do Not Refresh" - these are digital serials that should be linked to a catkey but should not be refreshed from source.

mjgiarlo:
@arcadiafalcone 💬 
> @mjgiarlo @justinlittman One exception is objects with the tag "Warning : Do Not Refresh" - these are digital serials that should be linked to a catkey but should not be refreshed from source.

so for those objects, should we also throw a cocina-models error if any catkeys have refresh set to true?

mjgiarlo:
This will be accommodated in https://github.com/sul-dlss/dor-services-app/issues/3767

PR #396, opened 2022-04-12
Generate markdown from descriptive types

### LINKED ISSUE #385: Generate user documentation from description validation files
In order to prevent having to maintain documentation of evolving validation standards in multiple places, it would be useful to be able to automatically generate user-readable documentation directly from the YAML configuration file for type validation.

Some notes on the conversion:
1. The section header in the YAML file is converted to separate words + "types", so contributor.name becomes "Contributor name types"
2. If the section header ends with structuredValue, "part types for structured value" is added in the markdown, so contributor.name.structuredValue becomes "Contributor name part types for structuredValue"
3. If the section header ends with groupedValue, "types for grouped value (MODS legacy)" is added in the markdown, so contributor.name.groupedValue becomes "Contributor name types for grouped value (MODS legacy)"

mjgiarlo:
Fixes #385

This commit adds the ability to generate user-facing documentation about how Cocina validates descriptive types. It is documented in the README and rendered by redoc via being linked to the OpenAPI specification.

CI

### DIFF
--- a/README.md
+++ b/README.md
@@ -31,6 +31,11 @@ exe/generator generate_schema DRO
 exe/generator generate_vocab
 ```
 
+### Descriptive documentation
+```
+exe/generator generate_descriptive_docs
+```
+
 ## Testing
 
 The generator is tested via its output when run against `openapi.yml`, viz., the Cocina model classes. Thus, `generate` should be run after any changes to `openapi.yml`.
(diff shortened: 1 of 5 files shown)


mjgiarlo:
unfortunately, our current openapi validation does not allow a markdown file as a ref. Worth looking into whether redoc/redocly has different tooling? Or we can go with a different implementation that doesn't involve the openapi/redocly infrastructure.

mjgiarlo:
@justinlittman Even better. Thanks for the assist!

PR #399, opened 2022-04-13
Adds purl validation.

### LINKED ISSUE #397: PURL validation
Validate that the bare identifier in the purl (description > purl) matches the bare identifier of the external identifier.

For example, `https://purl.stanford.edu/dd999df4567` and `druid:dd999df4567` match.

When mismatched, report "Purl mismatch: #{druid} purl does not match object druid."

justinlittman:
closes #397

Valid data > invalid data

### DIFF
--- a/lib/cocina/models/validators/purl_validator.rb
+++ b/lib/cocina/models/validators/purl_validator.rb
@@ -1,0 +1,51 @@
+# frozen_string_literal: true
+
+module Cocina
+  module Models
+    module Validators
+      # Validates that Purl matches the external identifier (druid)
+      class PurlValidator
+        def self.validate(clazz, attributes)
+          new(clazz, attributes).validate
+        end
+
+        def initialize(clazz, attributes)
+          @clazz = clazz
+          @attributes = attributes
+        end
+
+        def validate
+          return unless meets_preconditions?
+
+          return if identifier_from_druid == identifier_from_purl
+
+          raise ValidationError, "Purl mismatch: #{druid} purl does not match object druid."
+        end
+
+        private
+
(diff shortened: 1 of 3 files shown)


lib/cocina/models/validators/purl_validator.rb
```diff
@@ -0,0 +1,51 @@
...
+        def purl
+          @purl ||= attributes.dig(:description, :purl)
+        end
+
+        def identifier_from_druid
+          druid.delete_prefix('druid:')
+        end
+
+        def identifier_from_purl
+          purl[purl.rindex('/') + 1..]
```
jcoyne, line 46: I think `purl.split('/').last` might be easier to read and understand. 🤷

lib/cocina/models/validators/purl_validator.rb
```diff
@@ -0,0 +1,51 @@
...
+
+          raise ValidationError, "Purl mismatch: #{druid} purl does not match object druid."
+        end
+
+        private
+
+        attr_reader :clazz, :attributes
+
+        def meets_preconditions?
+          druid && purl
```
jcoyne, line 30: If it has a druid, then it is required to have a purl, right?  Do we need to check both?

justinlittman: Other way around: if it has a purl it must have a druid (e.g., for AdminPolicy that doesn't require description). Changing.

lib/cocina/models/validators/purl_validator.rb
justinlittman: Agree.

jcoyne (approved):

PR #403, opened 2022-04-13
Fix type validation when string keys.

justinlittman:

Descriptive type validation wasn't working when keys were strings.

Unit


spec/cocina/models/description_spec.rb
```diff
@@ -82,16 +82,6 @@
     end
   end
 
-  context 'with another etd' do
```
justinlittman, line 85: This wasn't adding any value.

jcoyne (approved):

spec/cocina/models/validators/description_types_validator_spec.rb
```diff
@@ -124,7 +124,7 @@
     it 'raises' do
       expect do
         validate
-      end.to raise_error(Cocina::Models::ValidationError, 'Unrecognized types in description: contributor1')
+      end.to raise_error(Cocina::Models::ValidationError, 'Unrecognized types in description: contributor1 (foo)')
```
jcoyne, line 127: way better message!

spec/cocina/models/validators/description_types_validator_spec.rb
```diff
@@ -441,4 +442,28 @@
...
+            ],
+            'type' => 'foo',
+            'status' => 'primary'
+          }
+        ]
+      }
+    end
+
+    it 'is not valid' do
+      expect { validate }.to raise_error(Cocina::Models::ValidationError)
```
edsu, line 466: Am I reading this right that `DescriptionTypesValidator.validate` will throw a validation exception if given attributes that have string keys, but that you can still instantiate an instance of `DescriptionTypesValidator` with attributes that have string keys?

justinlittman: `DescriptionTypesValidator.validate` will properly validate if given props regardless of whether keys are strings or symbols.

PR #406, opened 2022-04-14
Add validation of associated names. 

### LINKED ISSUE #391: Validate title note with type "associated name"
Create a new validator that validates that when title.note with type "associated name" has a value, it must match a contributor name.

When there is a mismatch, it should raise a validation error with the message "Missing data: Name associated with uniform title does not match any contributor."

refs https://github.com/sul-dlss/argo/issues/3375

justinlittman:
closes #391

* More validation
* Code re-use

### DIFF
--- a/lib/cocina/models/builders/name_title_group_builder.rb
+++ b/lib/cocina/models/builders/name_title_group_builder.rb
@@ -1,0 +1,134 @@
+# frozen_string_literal: true
+
+module Cocina
+  module Models
+    module Builders
+      # Helpers for MODS nameTitleGroups.
+      #   MODS titles need to know if they match a contributor and thus need a nameTitleGroup
+      #   MODS contributors need to know if they match a title and thus need a nameTitleGroup
+      #     If there is a match, the nameTitleGroup number has to be consistent for the matching title(s)
+      #     and the contributor(s)
+      class NameTitleGroupBuilder
+        # When to assign nameTitleGroup to MODS from cocina:
+        #   for cocina title of type "uniform",
+        #     look for cocina title properties :value or :structuredValue (recurse down through :parallelValue
+        #     as needed), and look for associated :note with :type of "associated name" at the level of the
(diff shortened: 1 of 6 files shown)


lib/cocina/generator/generator.rb
```diff
@@ -122,10 +122,7 @@ def generate_for(schema)
 
       NO_CLEAN = [
         'checkable.rb',
-        'dro_rights_description_builder.rb',
```
justinlittman, line 125: These were moved into own module, so no longer needed here.

lib/cocina/models/builders/dro_rights_description_builder.rb
```diff
@@ -0,0 +1,69 @@
+# frozen_string_literal: true
```
justinlittman, line 1: This is just a move.

lib/cocina/models/builders/name_title_group_builder.rb
```diff
@@ -0,0 +1,134 @@
+# frozen_string_literal: true
```
justinlittman, line 1: This is copied from DSA.

lib/cocina/models/builders/rights_description_builder.rb
```diff
@@ -0,0 +1,83 @@
+# frozen_string_literal: true
```
justinlittman, line 1: This is a move.

lib/cocina/models/builders/title_builder.rb
```diff
@@ -0,0 +1,213 @@
+# frozen_string_literal: true
```
justinlittman, line 1: This is a move.

jcoyne (changes requested):
This PR is very large and challenging to review. Can you do the moves in a separate PR?

justinlittman:
This was tested on sdr-deploy.

jcoyne (approved):

PR #417, opened 2022-04-15
Define the dry-struct so as to avoid metadata

jcoyne:

The old way made introspection impossible:

```
node = Cocina::Models::DescriptiveValue.schema.find { |key| key.name == :source }
node.type
node.type.right
#<Class:0x000000010ae7f710>
```

Added a unit test.

### DIFF
--- a/lib/cocina/generator/schema_base.rb
+++ b/lib/cocina/generator/schema_base.rb
@@ -23,14 +23,6 @@ module Cocina
         key || schema_doc.name
       end
 
-      # Allows non-required values to not be provided. This allows smaller
-      # requests as not every field needs to be present.
-      def omittable
-        return '' if required && !relaxed
-
-        '.meta(omittable: true)'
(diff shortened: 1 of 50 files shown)


justinlittman (commented):
Is this going to break everything?

jcoyne:
@justinlittman I don't know.  The tests here seem to pass.

justinlittman:
If we've learned anything, it is that the tests passing means nothing about our actual data.

Running a validation report. I'll let you know what it finds.

justinlittman:
Can you test branch against DSA and Argo?

jcoyne:
Both Argo (https://github.com/sul-dlss/argo/pull/3517) and DSA (https://github.com/sul-dlss/dor-services-app/pull/3889) don't seem to have problems with this change.

justinlittman (approved):

PR #419, opened 2022-04-15
Add validation for catalog link

jcoyne:
from https://github.com/sul-dlss/dor-services-app/blob/106cd76eff236b7bb823629d68b4da01706d11e4/app/services/metadata_service.rb#L8

Ref https://github.com/sul-dlss/argo/issues/3512

### DIFF
--- a/openapi.yml
+++ b/openapi.yml
@@ -383,6 +383,7 @@ components:
         catalogRecordId:
           description: Record identifier that is unique within the context of the linked record's catalog.
           type: string
+          pattern: '^\d+(:\d+)*$'
           example: '11403803'
       required:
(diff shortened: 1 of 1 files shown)


justinlittman (commented):
Have you checked if existing data satisfies this requirement?

jcoyne:
My assumption was that all were valid due to the validation linked to in the description, but just in case I made this report: https://github.com/sul-dlss/dor-services-app/pull/3887

jcoyne:
I examined 1.65 million records with identityMetadata and they didn't have any errors.

justinlittman (approved):

PR #422, opened 2022-04-18
[WIP] validate that cdl for specific conditions

### LINKED ISSUE #405: stanford/none access requires controlledDigitalLending to be present
For example: `{ view: 'stanford', download: 'none' }` would not be valid, but both of these are valid:
* `{ view: 'stanford', download: 'none', controlledDigitalLending: true }` 
* `{ view: 'stanford', download: 'none', controlledDigitalLending: false }` 

Ideally we should be able to omit controlledDigitalLending and it would default to false.

Make a proposal for how to model this.  Please mark any resulting PRs as "[HOLD]" and require round trip testing.

peetucket:

Ref #405 - validate the CDL attribute is present when access="stanford" and view="one"

TODO: can we set a default of false instead of throwing a validation error instead?  I'm not sure of the best way of doing this.

new tests

### DIFF
--- a/lib/cocina/models/validators/cdl_validator.rb
+++ b/lib/cocina/models/validators/cdl_validator.rb
@@ -1,0 +1,44 @@
+# frozen_string_literal: true
+
+module Cocina
+  module Models
+    module Validators
+      # Validates that the controlledDigitalLending attribute is set when view: 'stanford', download: 'none'
+      # though it will set a default of "false" if not found instead of throwing a validation error
+      class CdlValidator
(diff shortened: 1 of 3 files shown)


jcoyne:
@peetucket I think this may have missed the intention of #405.  We were looking for a better way to write the openapi.yml, not to add a new ruby validator.

peetucket:
Ok, so investigate if it's possible to do this type of validation via openapi specification instead of custom class(es) and implement if so.

PR #425, opened 2022-04-21
Update the README about camelCase etc.

mjgiarlo:


README.md
```diff
@@ -6,11 +6,11 @@
 
 # Cocina::Models
 
-This is a Ruby implementation of the SDR data model (named "COCINA"). The data being modeled includes digital repository objects.
+The cocina-models gem is a Ruby implementation of the Stanford Digital Repository (SDR) data model, which we named "Cocina." The data being modeled is oriented around digital repository objects.
```
ndushay, line 9: I would kill "oriented around" -- i think it adds nothing to the meaning.

jcoyne (approved):

ndushay (commented):
Mike,

Thanks for doing this.  I'm finding your prose a bit too indirect in a few spots and was in the process of suggesting the following when the PR was merged.

I would change what you wrote to:

The cocina-models gem is a Ruby implementation of the Stanford Digital Repository (SDR) data model, which we named "Cocina." The data being modeled are objects stored in the SDR.

We provide an OpenAPI specification of the cocina data-model in openapi.yml here.  This allows us to use community tools such as `OpenAPIParser` and `committee` gems for validation.  

This gem provides a set of generators (see below) which create Ruby classes from the OpenAPI specification, using modeling adhering to `dry-struct` and `dry-types` which allow for cocina consumers to validate SDR objects against the models here, and to manipulate the objects.

(your last paragraph is fine)

PR #427, opened 2022-04-21
Redesign descriptive types documentation

### LINKED ISSUE #415: Design updates for descriptive type documentation
The Markdown version of documentation at https://github.com/sul-dlss/cocina-models/blob/main/docs/description_types.md is closer to the desired display than the HTML version at https://sul-dlss.github.io/cocina-models/description_types.html, so the latter is not necessary.

1. The description should be on the line after the type term, with an indented bullet point.
2. Keep camelCase in headers.
3. If a type term has a status and/or use attribute, put it in an indented bullet point after the description (if present) in this form: `Deprecated. Preferred usage: #{value of use attribute}.`. Not all deprecated values will have a use attribute.
4. Add `.type` to the end of each path.
5. Don't include "Description types" as the top level header.
6. The header size should go down one each time a part is added to the path. So "Contributor types" (contributor.type) would be header 1, "Contributor name types" (contributor.name.type) would be header 2, and "Contributor name part types for structuredValue" (contributor.name.structuredValue.type) would be header 3.

See https://github.com/sul-dlss/cocina-models/blob/desc-documentation-sample/docs/descMD/documentation_test.md for an example of the desired formatting.

mjgiarlo:

Fixes #415

CI

### DIFF
--- a/docs/_config.yml
+++ b/docs/_config.yml
@@ -1,1 +1,0 @@
-theme: jekyll-theme-tactile
--- a/docs/description_types.md
+++ b/docs/description_types.md
@@ -1,170 +1,269 @@
-# Description types
-
-## Access accesscontact types
-_Path: access.accessContact_
-  * email: Email address for a contact person or institution concerning the resource.
-  * repository: Institution providing access to the resource.
-
-## Access digitallocation types
-_Path: access.digitalLocation_
-  * discovery: Online location for the purpose of discovering the resource.
-
+## Access accessContact types
+_Path: access.accessContact.type_
+  * email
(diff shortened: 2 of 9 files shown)


mjgiarlo:
@arcadiafalcone Here's the resulting raw markdown: https://raw.githubusercontent.com/sul-dlss/cocina-models/84e3d4cfe97c57f05b70794d776847b60e84dcc5/docs/description_types.md

Thanks!

justinlittman (approved):

arcadiafalcone:
Looks great!

arcadiafalcone (approved):

PR #428, opened 2022-04-22
Import the MODS to cocina descriptive translations

### LINKED ISSUE #426: Move Cocina descriptive / MODS mapping
refs https://github.com/sul-dlss/argo/issues/3551

Cocina descriptive / MODS mapping should be extracted from DSA and moved here. This will allow it to be utilized in DSA and Argo.

jcoyne:
So we can distribute this code.  Namely to Argo so it can generate cocina from modsulator and we can retire the legacy API.
Fixes #426

Tested in DSA, see https://github.com/sul-dlss/dor-services-app/pull/3917


justinlittman (changes requested):
This also requires the `spec/services/cocina/mapping/descriptive` specs.

How about making `Descriptive` a builder?

jcoyne:
@justinlittman  a builder?

justinlittman:
`Cocina::Models::Builders::DescriptiveBuilder` instead of `Cocina::Models::ToFedora/FromFedora::Descriptive`

jcoyne:
It's not really a Builder pattern though. The builder is a specific thing: https://github.com/davidgf/design-patterns-in-ruby/blob/master/builder.md

jcoyne:
@justinlittman I've added the missing specs.  Let's just get this PR in as minimally changed from DSA.  Then I want to make some changes as a follow up.  I don't think that `ToFedora` or `FromFedora` are appropriate names anymore. Maybe `ToMODS` and `FromMODS` instead?

justinlittman:
I don't see the new specs:
image

What am I missing?

jcoyne:
@justinlittman I see them. Refresh?
Screen Shot 2022-04-22 at 10 17 07 AM

jcoyne:
@justinlittman it's inside the from_fedora and to_fedora directory, which you have collapsed.

justinlittman (approved):

ndushay:
The "fixes #2426" refers to ... nothing.   Is this related to sul-dlss/argo/issues/3551 as the original issue?

jcoyne:
@ndushay that was supposed to be #426. Thanks for catching that.

ndushay (approved):
h2 access, h2_admin_metadata -> Added to renaming issue #431.

Should we include the datacite mapping as they are just as static as MODS, and in the same vein?  (we included mods/datacite_extension)   (As separate ticket/PR)

Approving, but please consider datacite mappings (could be a non-priority ticket)

PR #443, opened 2022-04-27
Adds validation of description values.

### LINKED ISSUE #424: Validation: only one of value, parallelValue, groupedValue, and structuredValue should be present
On the same level an object should have either a value or a structuredValue, not both.

justinlittman:
closes #424

Per Arcadia.

Unit

### DIFF
--- a/lib/cocina/models/validators/description_values_validator.rb
+++ b/lib/cocina/models/validators/description_values_validator.rb
@@ -1,0 +1,77 @@
+# frozen_string_literal: true
+
+module Cocina
+  module Models
+    module Validators
+      # Validates that there is only one of value, groupedValue, structuredValue, or parallelValue.
+      class DescriptionValuesValidator
+        def self.validate(clazz, attributes)
+          new(clazz, attributes).validate
+        end
+
+        def initialize(clazz, attributes)
+          @clazz = clazz
+          @attributes = attributes.deep_symbolize_keys
+          @error_paths = []
+        end
+
+        def validate
+          return unless meets_preconditions?
+
+          validate_obj(attributes, [])
+
+          return if error_paths.empty?
+
+          raise ValidationError, "Multiple value, groupedValue, structuredValue, and parallelValue in description: #{error_paths.join(', ')}"
+        end
+
+        private
+
+        attr_reader :clazz, :attributes, :error_paths
+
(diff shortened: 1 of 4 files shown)


justinlittman:
This is going to have to be held until after migration for testing against prod.

justinlittman:
Tested against QA, stage, and prod with no issues.

spec/cocina/models/validators/description_values_validator_spec.rb
```diff
@@ -0,0 +1,128 @@
...
+
+    it 'does not raise' do
+      validate
+    end
+  end
+
+  describe 'when an invalid Description' do
+    # let(:clazz) { Cocina::Models::RequestDRO }
+    let(:props) { invalid_desc_props }
+    # let(:dro_props) { { description: invalid_desc_props } }
```
mjgiarlo, line 100: Are the commented-out variables cruft?

lib/cocina/models/validators/description_values_validator.rb
```diff
@@ -0,0 +1,77 @@
...
+          end
+        end
+
+        def validate_obj(obj, path)
+          validate_hash(obj, path) if obj.is_a?(Hash)
+          validate_array(obj, path) if obj.is_a?(Array)
+        end
+
+        def validate_values(hash, path)
+          return unless hash.to_a.filter { |key, value| %i[value groupedValue structuredValue parallelValue].include?(key) && value.present? }.size > 1
```
mjgiarlo, line 56: How would you feel about using `Hash#count` for this? So, like:

```suggestion
  return if hash.count { |key, value| %i[value groupedValue structuredValue parallelValue].include?(key) && value.present? } < 2
```

lib/cocina/models/validators/description_values_validator.rb
```diff
@@ -0,0 +1,77 @@
...
+
+          error_paths << path_to_s(path).to_s
```
mjgiarlo, line 58: Do you need the `#to_s` on the end? I assumed based on the method name that `#path_to_s` might return a string.

justinlittman:
Thanks @mjgiarlo. All changes made.

mjgiarlo (approved):

PR #446, opened 2022-04-27
Handle access defaults in DarkValidator.

justinlittman:

The Dark Validator wasn't taking into account defaults. (That is: `DROAccess.new(view: 'dark') == DROAccess.new`).

Unit


ndushay (approved):

andrewjbtw:
What does this do?

justinlittman:
@andrewjbtw From your perspective, hopefully nothing. However, it plugs a possible hole in the DarkValidator (which there is no reason to think was ever a problem except in tests).

Issue #431, opened 2022-03-14
Refactor:  rename mapping folders "to_mods" and "from_mods", etc.


ndushay:
Since mapping to and from MODS is going to be with us for a while, but Fedora is going away, we should change the name of these two folders in app/services/cocina 

There is also inconsistency in class names:

should it always be name for the MODS element?  for the Cocina class?  For the originating object?    We have, for example, "contributor" and "name_builder" classes in the from_fedora mods code.

justinlittman:
This should be under `Cocina::Models::Mappings` ...

ndushay:
the h2 access, admin_metadata mappings:  files need renaming to correspond to mods...

ndushay:
And ... this is now done.

PR #449, opened 2022-04-27
refactor: mods mapping classes no longer nested inside Descriptive class

ndushay:

MODS maps to more than description in cocina; it also maps to access and admin.  Also, fewer nested classes save lives, or at least some typing.

part of #431

No openapi.yml was harmed in the creation of this PR.

specs pass.


justinlittman (approved):

jcoyne:
I sort of liked the Descriptive namespace here because it was indicating that we are mapping to/from `Cocina::Models::Description` and not from the whole object (e.g. `Cocina::Models::DRO`)

PR #473, opened 2022-05-06
Validation performance improvements

justinlittman:

Improve performance by stopping redundant validation.

Before:
```
puts Benchmark.measure { Cocina::Models::Validators::Validator.validate(c.class, c) }
  1.381309   0.008735   1.390044 (  1.393821)
```

After:
```
puts Benchmark.measure { Cocina::Models::Validators::Validator.validate(c.class, c) }
  0.588038   0.007669   0.595707 (  0.596772)
```

### DIFF
--- a/spec/cocina/models/validators/associated_name_validator_spec.rb
+++ b/spec/cocina/models/validators/associated_name_validator_spec.rb
@@ -3,8 +3,6 @@
 require 'spec_helper'
 
 RSpec.describe Cocina::Models::Validators::AssociatedNameValidator do
-  let(:validate) { described_class.validate(clazz, props) }
-
   let(:clazz) { Cocina::Models::Description }
 
   let(:desc_props) do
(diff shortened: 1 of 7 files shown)


spec/cocina/models/validators/associated_name_validator_spec.rb
```diff
@@ -149,25 +149,15 @@
...
     let(:clazz) { Cocina::Models::DRO }
 
     let(:contributor_name) { 'not Smith, John' }
 
     let(:props) { { description: desc_props } }
 
-    it 'raises' do
-      expect { validate }.to raise_error(Cocina::Models::ValidationError)
+    it 'does not validate' do
+      expect { validate }.not_to raise_error(Cocina::Models::ValidationError)
```
edsu, line 160: If the test for 'does not validate' why does it not expect a validation error here?

justinlittman: I agree this is unclear. Refactoring tests.

edsu (approved):
Thanks for clarifying the tests, and adding the bechmark!

PR #478, opened 2022-05-13
Retain role in name subject in Cocina>MODS direction

### LINKED ISSUE #465: Retain role in name subject in Cocina>MODS direction
Example of role being dropped: compare Cocina and published purl MODS for name subject "Brandin, Alf" in dv559gn8984.

See spec at https://github.com/sul-dlss/cocina-models/blob/9c9c0ce7f9e8002aa1e80233d1a4741f0a7e0a06/spec/cocina/models/mapping/descriptive/mods/subject_name_spec.rb#L809

justinlittman:
closes #465

Per Arcadia.

### DIFF
--- a/lib/cocina/models/mapping/to_mods/subject.rb
+++ b/lib/cocina/models/mapping/to_mods/subject.rb
@@ -394,7 +394,7 @@ module Cocina
             xml.name name_attrs do
               write_name_part(subject_value)
               write_display_form(display_values)
-              write_roles(subject.note)
+              write_roles(subject_value.note)
               write_other_notes(subject.note, 'description')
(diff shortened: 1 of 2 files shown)


edsu (approved):
I'm just double checking, does the call to `write_roles()` call on [line 411](https://github.com/sul-dlss/cocina-models/blob/93e3f9994738b673cf80aa430870601daa3584ce/lib/cocina/models/mapping/to_mods/subject.rb#L411) need to use `subject_value` as well?

justinlittman:
Great question, @edsu. When I was working on the ticket, I changed that as well but it broke some tests so apparently not.

PR #484, opened 2022-06-08
Validate descriptive dates per specified code

### LINKED ISSUE #392: Date validation
From sul-dlss/argo#3375, validates dates (per @arcadiafalcone )
```
Date encodings

w3cdtf, edtf, iso8601
Year:
YYYY (eg 1997)
Year and month:
YYYY-MM (eg 1997-07)
Complete date:
YYYY-MM-DD (eg 1997-07-16)
Complete date plus hours and minutes:
YYYY-MM-DDThh:mm (eg 1997-07-16T19:20)
Complete date plus hours, minutes and seconds:
YYYY-MM-DDThh:mm:ss (eg 1997-07-16T19:20:30)
Complete date plus hours, minutes, seconds and a decimal fraction of a second
YYYY-MM-DDThh:mm:ss.s (eg 1997-07-16T19:20:30.45)
Complete date plus hours and minutes with time zone:
YYYY-MM-DDThh:mmTZD (eg 1997-07-16T19:20+01:00)
Complete date plus hours, minutes and seconds:
YYYY-MM-DDThh:mm:ssTZD (eg 1997-07-16T19:20:30+01:00)
Complete date plus hours, minutes, seconds and a decimal fraction of a second
YYYY-MM-DDThh:mm:ss.sTZD (eg 1997-07-16T19:20:30.45+01:00)

edtf
-YYYY (eg -3999)

iso8601
YYYYMMDD
YYYYMMDDThhmm
YYYYMMDDThhmmss
YYYYMMDDhhmm
YYYYMMDDhhmmss
YYYYMMDDThhmmss.s+
YYYYMMDDhhmmss.s+
```

### LINKED ISSUE #483: copy date validator to cocina models from dor-services-app
I believe it is 2 files in DSA:
- app/validators/date_time_validator.rb
- spec/validators/date_time_validator_spec.rb

These should be copied, with the improvements from any remaining DSA work pertaining to the reports asked for by:
- sul-dlss/dor-services-app#4071
- sul-dlss/dor-services-app#4072
- sul-dlss/dor-services-app#4073

Afterwards:
See PR sul-dlss/dor-services-app#4052: take off Hold when this ticket is done

See also #392, which is, I believe, an older version of this.

NOTE:  you must ask @arcadiafalcone if this validation should be hooked up yet, or if that should be ticketed separately.

mjgiarlo:

Fixes https://github.com/sul-dlss/cocina-models/issues/392
Fixes https://github.com/sul-dlss/cocina-models/issues/483

This commit adds a new validator, but does not yet hook it up (follow-on work pending confirmation from Arcadia and Andrew), to make sure EDTF, ISO8601, and W3CDTF dates are valid per their specifications. To do the validation, we use upstream libraries in favor of rolling our own regexes or parsers because dragons, maybe.

One exception: the upstream W3CDTF parser flags two patterns we care about as invalid but per the spec should be valid. I briefly considered patching the library in question but this code has not been substantially touched in 10-15 years and I am loath to dig into it. In lieu of that, I added those cases in a one-liner that has code comments.

The tests written are currently incomplete, as I want to include some complex, nested Cocina structures to make sure the validator works against basic values, structured values, etc.

CI


lib/cocina/models/validators/date_time_validator.rb
```diff
@@ -0,0 +1,102 @@
...
+          Time.w3cdtf(value)
+          true
+        rescue StandardError
+          # NOTE: the upstream W3CDTF implementation in the `rss` gem does not
+          #       allow two patterns that should be valid per the specification:
+          #
+          # * YYYY
+          # * YYYY-MM
+          #
+          # So we catch the false positives from the upstream gem and allow
```
ndushay, line 91: aren't these false negatives?   As in they don't pass, but they should?

ndushay (approved):
Mike, this looks grand.

One comment quibble for your consideration.

Sad to see rubocop exceptions going in the wrong direction, but prolly no one regenned that file in a while.

lib/cocina/models/validators/date_time_validator.rb
mjgiarlo: Maybe? Depends on how you look at it. I was thinking false positives because they are incorrectly flagged as invalid.

ndushay: whatevs, not crucial at all.

PR #492, opened 2022-07-05
Handle missing value when mapping geo to mods.

### LINKED ISSUE #491: Dry::Struct::MissingAttributeError: Missing attribute: :value
## Backtrace

    line 16 of [PROJECT_ROOT]/app/services/publish/public_desc_metadata_service.rb: doc
    line 106 of [PROJECT_ROOT]/app/services/publish/public_desc_metadata_service.rb: remove_related_item_nodes_for_collections!
    line 97 of [PROJECT_ROOT]/app/services/publish/public_desc_metadata_service.rb: add_collection_reference!

[View full backtrace and more info at honeybadger.io](https://app.honeybadger.io/projects/50568/faults/86534060)

justinlittman:
closes #491

Publishing was failing when mapping to MODs.

Unit

### DIFF
--- a/lib/cocina/models/mapping/to_mods/geographic.rb
+++ b/lib/cocina/models/mapping/to_mods/geographic.rb
@@ -70,12 +70,10 @@ module Cocina
           end
 
           def extract_type(geo)
-            type = geo[:form].find do |form|
-              form[:type].match(TYPE_REGEX) || (form[:type].match(MEDIA_REGEX) && form[:value] == 'Image')
+            type = geo.form.find do |form|
+              form.type.match(TYPE_REGEX) || (form.type.match(MEDIA_REGEX) && form.value == 'Image')
             end
-            return type[:value] if type
-
-            nil
+            type&.value
           end
 
           def about(druid)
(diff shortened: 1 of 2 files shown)


lib/cocina/models/mapping/to_mods/geographic.rb
```diff
@@ -70,12 +70,10 @@ def extract_format(geo)
           end
 
           def extract_type(geo)
-            type = geo[:form].find do |form|
-              form[:type].match(TYPE_REGEX) || (form[:type].match(MEDIA_REGEX) && form[:value] == 'Image')
+            type = geo.form.find do |form|
+              form.type.match(TYPE_REGEX) || (form.type.match(MEDIA_REGEX) && form.value == 'Image')
```
mjgiarlo, line 74: Since we're not using the return value other than as a boolean, can we use `#match?` instead of `#match` here?

mjgiarlo (approved):
Approve pending consideration of admittedly non-major question

lib/cocina/models/mapping/to_mods/geographic.rb
justinlittman: Thanks. Fixed.

PR #494, opened 2022-07-09
Simplify Zeitwerk configuration

fxn:
Hi! I just saw this configuration in passing.

You can ignore the very `lib/cocina/rspec` directory. This is simpler and more performant than a wildcard pattern.


jcoyne (approved):
Thank you!

PR #508, opened 2022-08-04
Expose access to the druid regex

jcoyne:

This will enable H2 to decouple from druid-tools: https://github.com/sul-dlss/happy-heron/blob/dbf2991b058bb668e74c26b00d49d287ac920882/lib/tasks/deposit.rake#L45

ci

### DIFF
--- a/lib/cocina/models.rb
+++ b/lib/cocina/models.rb
@@ -153,5 +153,12 @@ module Cocina
       raise ValidationError, 'Type field not found'
     end
     private_class_method :type_for
+
+    def self.druid_regex
+      @druid_regex ||= begin
+        str = Openapi3Parser.load_file('openapi.yml').components.schemas['Druid'].pattern
+        Regexp.new(str)
+      end
+    end
   end
 end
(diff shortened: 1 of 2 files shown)


lib/cocina/models.rb
```diff
@@ -153,5 +153,12 @@ def self.type_for(dyn)
       raise ValidationError, 'Type field not found'
     end
     private_class_method :type_for
+
+    def self.druid_regex
+      @druid_regex ||= begin
+        str = Openapi3Parser.load_file('openapi.yml').components.schemas['Druid'].pattern
```
justinlittman, line 159: Are you sure the overhead of parsing openapi is worth it vs. just duplicating it here?

justinlittman (approved):
Approve pending consideration of comment.

lib/cocina/models.rb
jcoyne: Its memorized, so it should only be once per process. We can optimize it later if necessary.

ndushay: does openapi.yml strictly parse druids so the letters do not have vowels or the letter L ?

justinlittman: https://github.com/sul-dlss/cocina-models/blob/main/openapi.yml#L1045

Issue #512, opened 2022-08-17
Add DOI pattern exceptions


jcoyne:
These druids already have DOIs, but they can't be persisted.  Update the pattern to handle these:

hr045dx8661 | 10.25740/VA90-CT15
wv848yd6100 | 10.25740/syxa-m256
mp388hh4436 | 10.25740/12qf-5243
wd199pm1314 | 10.25740/65j8-6114
vx915bx0661 | 10.25936/629T-BX79

Encode these as exceptions, not a general pattern. as we are committed not to having further exceptions.

ndushay:
- cocina-models PR merged
- cocina-models 0.84.0 released
- sdr-client, dor-services-client PRs to use cocina-models 0.84.0 merged
- sdr-client, dor-services-client gems new version released
- dor-services-app, sdr-api PRs with updated openapi.yml  and gem updates (cocina-models, etc.) released
- cocina level 2 PRs created w jenkins
- cocina level 2 PRs merged with access-update-scripts
- cocina apps tagged with rel-2022-08-23
- cocina apps deployed to stage and integration tests run until they passed.

PR #514, opened 2022-08-23
DOI schema gets exceptions for 5 hydrus identiiers

ndushay:

Adds DOI exceptions (from H2 work) to the DOI schema.

Part of #512

rubular

### DIFF
--- a/openapi.yml
+++ b/openapi.yml
@@ -902,9 +902,9 @@ components:
         - $ref: '#/components/schemas/DoiExceptions'
     DoiExceptions:
       type: string
-      description: pre-existing Digital Object Identifiers (https://www.doi.org) not matching the pattern
-      pattern: '^10\.(25740\/(VA90-CT15|syxa-m256|12qf-5243|65j8-6114)|25936\/629T-BX79)$'
-      example: '10.25740/12qf-5243'
+      description: pre-existing Digital Object Identifiers (https://www.doi.org) not matching the pattern (case insensitive)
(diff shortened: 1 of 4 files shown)


openapi.yml
```diff
@@ -896,8 +896,18 @@ components:
               # description: An alphabet or other notation used to represent a
               #   language or other symbolic system of the descriptive element value.
     DOI:
-      type: string
       description: Digital Object Identifier (https://www.doi.org)
+      oneOf:
+        - $ref: '#/components/schemas/DOI_pattern'
+        - $ref: '#/components/schemas/DOI_exceptions'
+    DOI_exceptions:
```
jcoyne, line 903: I think this will have to be DoiExceptions, so that zeitwork can predict where this file is.

openapi.yml
```diff
@@ -896,8 +896,18 @@ components:
...
+      type: string
+      description: pre-existing Digital Object Identifiers (https://www.doi.org) not matching the pattern
+      pattern: '^10\.(25740\/(VA90-CT15|syxa-m256|12qf-5243|65j8-6114)|25936\/629T-BX79)$'
```
justinlittman, line 906: Would doing this as an `enum` instead of a `pattern` be clearer?

openapi.yml
ndushay: yep - got that error and fixed it.

openapi.yml
ndushay: good idea.  I will change.

ndushay: answer:  DOI is NOT case sensitive, so pattern is better.

justinlittman (approved):

PR #521, opened 2022-09-01
Add edtf exceptions

jcoyne:
This will allow us to fix bad dates in the repository

ci

### DIFF
--- a/lib/cocina/models/validators/date_time_validator.rb
+++ b/lib/cocina/models/validators/date_time_validator.rb
@@ -76,7 +76,8 @@ module Cocina
           #
           # So we catch the false positives from the upstream gem and allow
           # this pattern to validate
-          /\AY-?\d{5,}\Z/.match?(value)
+          /\AY-?\d{5,}\Z/.match?(value) ||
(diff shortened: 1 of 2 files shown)


lib/cocina/models/validators/date_time_validator.rb
```diff
@@ -76,7 +76,8 @@ def valid_edtf?(value)
           #
           # So we catch the false positives from the upstream gem and allow
           # this pattern to validate
-          /\AY-?\d{5,}\Z/.match?(value)
+          /\AY-?\d{5,}\Z/.match?(value) ||
+            /\A\-?\d{1,3}\Z/.match?(value) # temporarily allow format violations
         end
```
peetucket, line 81: i'm not positive i fully understand this regex, shouldn't we allow 4 digits too?

jcoyne: 4 digits is the spec, not an exception.

peetucket: ahh, of course

peetucket (approved):

Issue #532, opened 2022-09-22
remove deprecated DescriptiveParallelContributor model (?)


ndushay:
DescriptiveParallelContributor class is marked deprecated:  https://github.com/sul-dlss/cocina-models/blob/main/lib/cocina/models/descriptive_parallel_contributor.rb#L5

https://github.com/sul-dlss/cocina-models/blob/main/openapi.yml#552

Can it be removed?

Arcadia says "parallelContributor was never in use"

Steps
- [ ] write a report to ensure it's not used (if it is, have Arcadia change those objects)
- [ ] take it out of the API and the code (by regenerating models) ... and go through the cocina gem release process.

ndushay:
no occurences found in DROs or Collections, using property_existence*.rb reports.

ndushay:
Correction:  there are lots of EMPTY occurrences of parallelContributor ...

justinlittman:
So does that mean this is un-removable?

jcoyne:
@justinlittman I think we could remove it by making it optional, omitable, and changing the default to nil. Then we can rewrite any/all descriptive metadata that has `parallelContributor: []`

ndushay:
I don't think the cost of always putting in empty arrays should be "we can never omit an array property from the model."

ndushay:
reran reports today;  still no non-empty values for parallelContributor for collections or DROs.   starting work on PR to 

(from JCoyne:)

> I think we could remove it by making it optional, omitable, and changing the default to nil. Then we can rewrite any/all descriptive metadata that has `parallelContributor: []`

PR #533, opened 2022-09-22
remove unused, deprecated parallelContributor

ndushay:

Simplify, simplify, simplify.  Arcadia has indicated that parallelContributor can go away.

I have verified with reports that parallelContributor is not used in any of our objects.

Part of #532


justinlittman:
Naive question: Have you verified with @arcadiafalcone that she doesn't intend to use parallelContributor in the future? (That is, it is speculative not deprecated.)

ndushay:
> Naive question: Have you verified with @arcadiafalcone that she doesn't intend to use parallelContributor in the future? (That is, it is speculative not deprecated.)

@justinlittman this is all with Arcadia's approval.

justinlittman (approved):

PR #564, opened 2023-02-07
Add catalog links for Folio

### LINKED ISSUE #561: Update CatalogLink definitions in cocina for FOLIO
The catalogLink.catalog enumerated list should also include `folio` and `previous folio`.

The regex in catalogLink.catalogRecordId should apply only when the catalog is `symphony` or `previous symphony`. When the record is from FOLIO, it should match the patterns for both migrated and new instance HRIDs.
Migrated from Symphony: `/^a\d+$/`
Migrated from Voyager: `/^L\d+$/`
Created in FOLIO: `/^in\d+$/` (to be confirmed)

Presumably editing openapi.yml is sufficient(?)

mjgiarlo:

Connects to #561

Includes:
* Turn CatalogLink model into a union type
  * Allows either Symphony-type catalog links ("catkeys" or "ckeys") or Folio-type catalog links ("HRIDs"). This split allows each type of catalog link to validate its own `catalog` string and `catalogRecordId` regex properties. 
* Discontinue allowing non-number characters in catkeys
  * Tightens the regular expression governing what is a valid Symphony catalog record ID. Arcadia confirms this is the right value, and it is aligned with the regex for the new Folio catalog links (of the migrated-from-symphony variety). Additionally, I ran a report in DSA (https://github.com/sul-dlss/dor-services-app/pull/4357) confirming there are zero catkeys in QA, stage, and prod that contain non-number characters.
* Regenerate all models.
  
CI, and tested in QA, stage, and prod envs on `sdr-infra`

### DIFF
--- a/lib/cocina/generator/schema_base.rb
+++ b/lib/cocina/generator/schema_base.rb
@@ -63,8 +63,8 @@ module Cocina
       def dry_datatype(doc)
         return doc.name if doc.name.present? && schemas.include?(doc.name)
 
-        datatype_from_doc_type(doc) ||
-          datatype_from_doc_names(doc) ||
+        datatype_from_doc_names(doc) ||
+          datatype_from_doc_type(doc) ||
           raise("#{doc.type} not supported")
       end
 
@@ -84,19 +84,17 @@ module Cocina
       end
 
       def datatype_from_doc_names(doc)
-        if defined_datatypes?(doc)
-          doc.one_of.map(&:name).join(' | ')
-        elsif any_datatype?(doc)
-          'Types::Nominal::Any'
-        end
+        return unless defined_datatypes?(doc)
+
+        doc.one_of.map(&:name).join(' | ')
       end
 
       def defined_datatypes?(doc)
-        doc.one_of&.map(&:name).all? { |name| name.present? && schemas.include?(name) }
+        doc.one_of&.map(&:name)&.all? { |name| name.present? && schemas.include?(name) }
       end
 
       def any_datatype?(doc)
-        relaxed || doc.one_of&.map(&:type).all? { |o| %w[integer string].include?(o) }
+        relaxed || doc.one_of&.map(&:type)&.all? { |o| %w[integer string].include?(o) }
       end
 
       def string_dry_datatype(doc)
--- a/lib/cocina/models/catalog_link.rb
+++ b/lib/cocina/models/catalog_link.rb
@@ -2,15 +2,6 @@
 
 module Cocina
   module Models
-    class CatalogLink < Struct
-      # Catalog that is the source of the linked record.
-      # example: symphony
-      attribute :catalog, Types::Strict::String.enum('symphony', 'previous symphony')
-      # Only one of the catkeys should be designated for refreshing. This means that this key is the one used to pull metadata from the catalog if there is more than one key present.
-      attribute :refresh, Types::Strict::Bool.default(false)
-      # Record identifier that is unique within the context of the linked record's catalog.
-      # example: 11403803
-      attribute :catalogRecordId, Types::Strict::String.constrained(format: /^\d+(:\d+)*$/)
-    end
+    CatalogLink = FolioCatalogLink | SymphonyCatalogLink
   end
 end
--- a/lib/cocina/models/folio_catalog_link.rb
+++ b/lib/cocina/models/folio_catalog_link.rb
@@ -1,0 +1,16 @@
+# frozen_string_literal: true
+
+module Cocina
+  module Models
+    # A linkage between an object and a Folio catalog record
+    class FolioCatalogLink < Struct
+      # Catalog that is the source of the linked record.
+      # example: folio
+      attribute :catalog, Types::Strict::String.enum('folio', 'previous folio')
+      # Only one of the catkeys should be designated for refreshing. This means that this key is the one used to pull metadata from the catalog if there is more than one key present.
+      attribute :refresh, Types::Strict::Bool.default(false)
+      # Record identifier that is unique within the context of the linked record's catalog.
+      attribute :catalogRecordId, Types::Nominal::Any
+    end
+  end
+end
--- a/lib/cocina/models/folio_catalog_link.rb
+++ b/lib/cocina/models/folio_catalog_link.rb
@@ -10,7 +10,7 @@ module Cocina
       # Only one of the catkeys should be designated for refreshing. This means that this key is the one used to pull metadata from the catalog if there is more than one key present.
       attribute :refresh, Types::Strict::Bool.default(false)
       # Record identifier that is unique within the context of the linked record's catalog.
-      attribute :catalogRecordId, Types::Nominal::Any
+      attribute :catalogRecordId, MigratedFromSymphonyIdentifier | MigratedFromVoyagerIdentifier | CreatedInFolioIdentifier
     end
   end
 end
--- a/lib/cocina/models/symphony_catalog_link.rb
+++ b/lib/cocina/models/symphony_catalog_link.rb
@@ -1,0 +1,17 @@
+# frozen_string_literal: true
+
+module Cocina
+  module Models
+    # A linkage between an object and a Symphony catalog record
+    class SymphonyCatalogLink < Struct
+      # Catalog that is the source of the linked record.
+      # example: symphony
+      attribute :catalog, Types::Strict::String.enum('symphony', 'previous symphony')
+      # Only one of the catkeys should be designated for refreshing. This means that this key is the one used to pull metadata from the catalog if there is more than one key present.
+      attribute :refresh, Types::Strict::Bool.default(false)
+      # Record identifier that is unique within the context of the linked record's catalog.
+      # example: 11403803
+      attribute :catalogRecordId, Types::Strict::String.constrained(format: /^\d+(:\d+)*$/)
+    end
+  end
+end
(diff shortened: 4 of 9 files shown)


justinlittman:
Any chance we can split this up into something more manageable?

openapi.yml
```diff
@@ -363,7 +363,23 @@ components:
       type: string
       pattern: '^245[0-9]{8}$'
       example: '24503259768'
-    CatalogLink:
+    CreatedInFolioIdentifier:
```
justinlittman, line 366: Can these be mixed in alphabetically like rest of classes?

openapi.yml
```diff
@@ -363,7 +363,23 @@ components:
...
+      description: A record identifier created in Folio
+      type: string
+      pattern: '^in\d+$'
+      example: 'in11403803'
+    MigratedFromSymphonyIdentifier:
+      description: A record identifier migrated from Symphony
+      type: string
+      pattern: '^a\d+$'
+      example: 'a11403803'
```
justinlittman, line 375: It looks like existing Symphony identifiers can be `123:456`. Does this need to allow?

mjgiarlo:
@justinlittman 💬 
> Any chance we can split this up into something more manageable?

For sure. My intent with this draft PR was primarily to get feedback on the new approach to re-using custom types. I should have been clearer about that. In the meantime, here's a start:

https://github.com/sul-dlss/cocina-models/pull/565

openapi.yml
mjgiarlo: The current components aren't super consistent about this but I can toss a commit at this branch that (re-)alphabetizes the components, once I get feedback on the new approach to custom types and union types.

openapi.yml
lwrubel: @justinlittman I'm still learning about Symphony identifiers--could you point me to an example of one of those with the `:` in it?

mjgiarlo: @arcadiafalcone do you have thoughts on ☝🏻 ?

openapi.yml
ndushay: it would be good to check with Darsi on whether it's going to be "in" and stay that way in "perpetuity" with Folio

ndushay (commented):
I'm with JLitt - it's hard to figure out which file changes are about the approach you want us to vet for the ticket vs. which are about changes you made for other reasons.  Can you please split this or redo this so there is a single focus PR for what you want us to check?

I'm not sure we gain a whole lot from what you did vs. just having a regex allow "a" "L" or "in" before the digits for Folio regexp.  That's an @arcadiafalcone question, I think -- is it worth having the separate definitions in the API?  Will we ever want to hone in on one group and need a way to do it with the API definitions vs. the regex?

I'm open to either but feeling like KISS might be better here.

mjgiarlo:
@ndushay 💬 
> I'm with JLitt - it's hard to figure out which file changes are about the approach you want us to vet for the ticket vs. which are about changes you made for other reasons. Can you please split this or redo this so there is a single focus PR for what you want us to check?

I've already split out three PRs with a fourth coming soon 🕺🏻 
 
> I'm not sure we gain a whole lot from what you did vs. just having a regex allow "a" "L" or "in" before the digits for Folio regexp. That's an @arcadiafalcone question, I think -- is it worth having the separate definitions in the API? Will we ever want to hone in on one group and need a way to do it with the API definitions vs. the regex?

I figure better safe than sorry, lest we say "OK" to a catalog link that looks like folio in one way and like symphony in another way.

lib/cocina/generator/schema_base.rb
```diff
@@ -63,8 +63,8 @@ def relaxed_comment
       def dry_datatype(doc)
         return doc.name if doc.name.present? && schemas.include?(doc.name)
 
-        datatype_from_doc_type(doc) ||
-          datatype_from_doc_names(doc) ||
+        datatype_from_doc_names(doc) ||
```
mjgiarlo, line 66: This re-ordering lets us simplify the code in this class (see `#datatype_from_doc_names` below) and does not break or otherwise change models in unexpected ways.

lib/cocina/generator/schema_base.rb
```diff
@@ -84,19 +84,17 @@ def datatype_from_doc_type(doc)
       end
 
       def datatype_from_doc_names(doc)
-        if defined_datatypes?(doc)
-          doc.one_of.map(&:name).join(' | ')
-        elsif any_datatype?(doc)
-          'Types::Nominal::Any'
```
mjgiarlo, line 90: With this change, there's only one path to `Any` types.

lib/cocina/generator/schema_base.rb
```diff
@@ -84,19 +84,17 @@ def datatype_from_doc_type(doc)
...
       end
 
       def defined_datatypes?(doc)
-        doc.one_of&.map(&:name).all? { |name| name.present? && schemas.include?(name) }
+        doc.one_of&.map(&:name)&.all? { |name| name.present? && schemas.include?(name) }
       end
 
       def any_datatype?(doc)
-        relaxed || doc.one_of&.map(&:type).all? { |o| %w[integer string].include?(o) }
+        relaxed || doc.one_of&.map(&:type)&.all? { |o| %w[integer string].include?(o) }
```
mjgiarlo, line 97: With the above logic change, we need to be a little more careful with mapping over the oneOfs.

lib/cocina/models/catalog_link.rb
```diff
@@ -2,15 +2,6 @@
...
-      # Catalog that is the source of the linked record.
-      # example: symphony
-      attribute :catalog, Types::Strict::String.enum('symphony', 'previous symphony')
-      # Only one of the catkeys should be designated for refreshing. This means that this key is the one used to pull metadata from the catalog if there is more than one key present.
-      attribute :refresh, Types::Strict::Bool.default(false)
-      # Record identifier that is unique within the context of the linked record's catalog.
-      # example: 11403803
-      attribute :catalogRecordId, Types::Strict::String.constrained(format: /^\d+(:\d+)*$/)
-    end
+    CatalogLink = FolioCatalogLink | SymphonyCatalogLink
```
mjgiarlo, line 5: It's now a union/sum type.

lib/cocina/models/folio_catalog_link.rb
```diff
@@ -0,0 +1,16 @@
...
+  module Models
+    # A linkage between an object and a Folio catalog record
+    class FolioCatalogLink < Struct
+      # Catalog that is the source of the linked record.
+      # example: folio
+      attribute :catalog, Types::Strict::String.enum('folio', 'previous folio')
+      # Only one of the catkeys should be designated for refreshing. This means that this key is the one used to pull metadata from the catalog if there is more than one key present.
+      attribute :refresh, Types::Strict::Bool.default(false)
+      # Record identifier that is unique within the context of the linked record's catalog.
+      attribute :catalogRecordId, MigratedFromSymphonyIdentifier | MigratedFromVoyagerIdentifier | CreatedInFolioIdentifier
```
mjgiarlo, line 13: Building on the past couple PRs, this PR can now use a union type of generated types.

lib/cocina/models/symphony_catalog_link.rb
```diff
@@ -0,0 +1,17 @@
...
+    # A linkage between an object and a Symphony catalog record
+    class SymphonyCatalogLink < Struct
+      # Catalog that is the source of the linked record.
+      # example: symphony
+      attribute :catalog, Types::Strict::String.enum('symphony', 'previous symphony')
+      # Only one of the catkeys should be designated for refreshing. This means that this key is the one used to pull metadata from the catalog if there is more than one key present.
+      attribute :refresh, Types::Strict::Bool.default(false)
+      # Record identifier that is unique within the context of the linked record's catalog.
+      # example: 11403803
+      attribute :catalogRecordId, Types::Strict::String.constrained(format: /^\d+$/)
```
mjgiarlo, line 14: This is the catkey regex change (see PR description for rationale and testing info).

openapi.yml
```diff
@@ -196,19 +196,6 @@ components:
...
-      properties:
-        hasAdminPolicy:
-          $ref: '#/components/schemas/Druid'
-        releaseTags:
-          description: Tags for release
-          type: array
-          items:
-            $ref: '#/components/schemas/ReleaseTag'
-      required:
-        - hasAdminPolicy
```
mjgiarlo, line 211: The two changes to the OpenAPI spec are:

1. Split catalog link schema into folio and symphony schemas (the meat of this PR); and
2. Alphabetize all the schemas (requested by @justinlittman in past round of code review).

mjgiarlo: I do not recommend trying to review the changes to this file. Suffice to say that the openapi validation in CI succeeds, the model changes generated from the spec are all as expected, and all the specs pass?

ndushay: is this due to alphabetization of the types?  Perhaps in a perfect world that would have been its own PR but I'll allow it.  :-D

ndushay (approved):
This is so much more tractable now that the refactor PRs were separated!

lib/cocina/generator/schema_base.rb
mjgiarlo: Generation is suuuuuper fast now.

ndushay: execution time, not generation time.  Like, when instantiating a cocina object from JSON or creating one from code or whatever.  Again ... prolly doesn't matter

mjgiarlo: AFAIK, this code is only called when generating.

justinlittman (approved):

PR #566, opened 2023-02-07
Generate models en masse and run rubocop en masse

mjgiarlo:

This commit speeds up model generation notably by significantly reducing the number of times we have to run rubocop, which is now down to one time.

CI + repeated generation runs

### DIFF
--- a/lib/cocina/generator/generator.rb
+++ b/lib/cocina/generator/generator.rb
@@ -19,14 +19,12 @@ module Cocina
       def generate
         clean_output
 
-        # rubocop:disable Style/HashEachMethods
-        # This is not a Hash
-        schemas.keys.each do |schema_name|
+        filepaths = schemas.keys.filter_map do |schema_name|
           schema = schema_for(schema_name)
           generate_for(schema) if schema
         end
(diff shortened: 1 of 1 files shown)


jcoyne (commented):
Does this still work even if you're only generating a single schema?

mjgiarlo:
@jcoyne 💬 
> Does this still work even if you're only generating a single schema (e.g. `exe/generator generate_schema DRO`)?

Yes, but that bit is jammed into another commit (pulling these small PRs out of another branch). I'll grab that. Good eye!

jcoyne (approved):

ndushay (approved):
this seems like an annoying weird hoop for using rubocop.

mjgiarlo:
@ndushay 💬 
> this seems like an annoying weird hoop for using rubocop.

On one hand, yeah. On the other, it's also nice to generate models w/o caring too much about indentation and whitespace and then fixing it all up in one fell swoop at the end.

justinlittman:
Yeah, the motivation behind the rubocop cleanup is that it was nearly impossible to get the indentation and whitespace correct when generating.

PR #568, opened 2023-02-08
Allow generators to create and use custom types and union types

mjgiarlo:

These commits change the Ruby models so that they are working more by reference and less by value, e.g., instead of having the druid regex in every model that has an `externalIdentifier`, point every `externalIdentifier` at the `Druid` model and let the model handle the validation.

Currently the models handle validation through a combination of validating against the OpenAPI spec and using the generated Ruby (dry-) types, and this commit adds a bit more smarts and descriptive oomph to the Ruby side.

A separate commit moves yet more smarts into the Ruby (dry-) types, allowing them to make use of union types defined in the OpenAPI spec, such as for barcodes and DOIs. This sets up a follow-on commit around Symphony and Folio catalog links which will similarly use union types.

CI & tested against QA, stage, and 100K prod records on `sdr-infra`

### DIFF
--- a/lib/cocina/generator/generator.rb
+++ b/lib/cocina/generator/generator.rb
@@ -5,7 +5,7 @@ require 'fileutils'
 module Cocina
   module Generator
     # Class for generating Cocina models from openapi.
-    class Generator < Thor
+    class Generator < Thor # rubocop:disable Metrics/ClassLength
       include Thor::Actions
 
       class_option :openapi, desc: 'Path of openapi.yml', default: 'openapi.yml'
@@ -100,9 +100,13 @@ module Cocina
 
         case schema_doc.type
         when 'object'
-          Schema.new(schema_doc)
+          Schema.new(schema_doc, schemas: schemas.keys)
         when 'string'
-          Datatype.new(schema_doc)
+          Datatype.new(schema_doc, schemas: schemas.keys)
+        when NilClass
+          return unless schema_doc.one_of.map(&:name).all? { |ref_schema_name| schemas.keys.include?(ref_schema_name) }
+
+          UnionType.new(schema_doc)
         end
       end
 
--- a/lib/cocina/generator/schema_array.rb
+++ b/lib/cocina/generator/schema_array.rb
@@ -9,7 +9,7 @@ module Cocina
       end
 
       def array_of_type
-        schema_doc.items.name || "Types::#{dry_datatype(schema_doc.items)}"
+        schema_doc.items.name || dry_datatype(schema_doc.items)
       end
     end
   end
--- a/lib/cocina/generator/schema_base.rb
+++ b/lib/cocina/generator/schema_base.rb
@@ -4,15 +4,16 @@ module Cocina
   module Generator
     # Base class for generating from openapi
     class SchemaBase
-      attr_reader :schema_doc, :key, :required, :nullable, :parent, :relaxed
+      attr_reader :schema_doc, :key, :required, :nullable, :parent, :relaxed, :schemas
 
-      def initialize(schema_doc, key: nil, required: false, nullable: false, parent: nil, relaxed: false)
+      def initialize(schema_doc, key: nil, required: false, nullable: false, parent: nil, relaxed: false, schemas: [])
         @schema_doc = schema_doc
         @key = key
         @required = required
         @nullable = nullable
         @parent = parent
         @relaxed = relaxed
+        @schemas = schemas
       end
 
       def filename
@@ -60,7 +61,7 @@ module Cocina
       end
 
       def dry_datatype(doc)
-        return doc.name if doc.name.present? && Cocina::Models.const_defined?(doc.name)
+        return doc.name if doc.name.present? && schemas.include?(doc.name)
 
         datatype_from_doc_type(doc) ||
           datatype_from_doc_names(doc) ||
@@ -91,7 +92,7 @@ module Cocina
       end
 
       def defined_datatypes?(doc)
-        doc.one_of&.map(&:name).all? { |name| name.present? && Cocina::Models.const_defined?(name) }
+        doc.one_of&.map(&:name).all? { |name| name.present? && schemas.include?(name) }
       end
 
       def any_datatype?(doc)
--- a/lib/cocina/generator/schema_base.rb
+++ b/lib/cocina/generator/schema_base.rb
@@ -60,18 +60,38 @@ module Cocina
       end
 
       def dry_datatype(doc)
+        return doc.name if doc.name.present? && Cocina::Models.const_defined?(doc.name)
+
+        datatype_from_doc_type(doc) ||
+          datatype_from_doc_names(doc) ||
+          raise("#{doc.type} not supported")
+      end
+
+      def datatype_from_doc_type(doc)
         case doc.type
         when 'integer'
-          'Strict::Integer'
+          'Types::Strict::Integer'
         when 'string'
           string_dry_datatype(doc)
         when 'boolean'
-          'Strict::Bool'
+          'Types::Strict::Bool'
         else
-          return 'Nominal::Any' if any_datatype?(doc)
+          return 'Types::Nominal::Any' if any_datatype?(doc)
 
           raise "#{schema_doc.type} not supported"
         end
+      end
+
+      def datatype_from_doc_names(doc)
+        if defined_datatypes?(doc)
+          doc.one_of.map(&:name).join(' | ')
+        elsif any_datatype?(doc)
+          'Types::Nominal::Any'
+        end
+      end
+
+      def defined_datatypes?(doc)
+        doc.one_of&.map(&:name).all? { |name| name.present? && Cocina::Models.const_defined?(name) }
       end
 
       def any_datatype?(doc)
@@ -79,12 +99,15 @@ module Cocina
       end
 
       def string_dry_datatype(doc)
-        case doc.format
-        when 'date-time'
-          'Params::DateTime'
-        else
-          'Strict::String'
-        end
+        format = case doc.format
+                 when 'date-time'
+                   'Types::Params::DateTime'
+                 else
+                   'Types::Strict::String'
+                 end
+        return format unless doc.pattern
+
+        "Types::Strict::String.constrained(format: /#{doc.pattern}/)"
       end
 
       def preamble
--- a/lib/cocina/generator/schema_value.rb
+++ b/lib/cocina/generator/schema_value.rb
@@ -16,7 +16,7 @@ module Cocina
       private
 
       def type
-        "Types::#{dry_datatype(schema_doc)}#{optional}#{default}#{enum}"
+        "#{dry_datatype(schema_doc)}#{optional}#{default}#{enum}"
       end
 
       def enum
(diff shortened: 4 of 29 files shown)


lib/cocina/generator/schema_array.rb
```diff
@@ -9,7 +9,7 @@ def generate
       end
 
       def array_of_type
-        schema_doc.items.name || "Types::#{dry_datatype(schema_doc.items)}"
+        schema_doc.items.name || dry_datatype(schema_doc.items)
```
mjgiarlo, line 12: The `Types::` prefix for built-in dry-types has moved below to avoid pre-pending it to custom/generated types, e.g., `Druid` or `SourceId`.

lib/cocina/generator/generator.rb
```diff
@@ -100,9 +100,13 @@ def schema_for(schema_name)
 
         case schema_doc.type
         when 'object'
-          Schema.new(schema_doc)
+          Schema.new(schema_doc, schemas: schemas.keys)
         when 'string'
-          Datatype.new(schema_doc)
+          Datatype.new(schema_doc, schemas: schemas.keys)
```
mjgiarlo, line 105: These changes allow all schemas to know the entirety of the named schemas from the API doc so that they can reference one another.

lib/cocina/generator/generator.rb
```diff
@@ -100,9 +100,13 @@ def schema_for(schema_name)
...
+        when NilClass
+          return unless schema_doc.one_of.map(&:name).all? { |ref_schema_name| schemas.keys.include?(ref_schema_name) }
+
+          UnionType.new(schema_doc)
```
mjgiarlo, line 109: This is what allows us to now generate Rubyland models for `Barcode` and `DOI`, for example.

lib/cocina/generator/schema_base.rb
```diff
@@ -60,31 +61,54 @@ def relaxed_comment
...
-          'Strict::Integer'
+          'Types::Strict::Integer'
         when 'string'
           string_dry_datatype(doc)
         when 'boolean'
-          'Strict::Bool'
+          'Types::Strict::Bool'
         else
-          return 'Nominal::Any' if any_datatype?(doc)
+          return 'Types::Nominal::Any' if any_datatype?(doc)
```
mjgiarlo, line 80: This is a bug I introduced in the prior PR earlier today while trying to break one PR into several. Fixing that here.

lib/cocina/generator/schema_base.rb
```diff
@@ -60,31 +61,54 @@ def relaxed_comment
...
-        end
+        format = case doc.format
+                 when 'date-time'
+                   'Types::Params::DateTime'
+                 else
+                   'Types::Strict::String'
+                 end
+        return format unless doc.pattern
+
+        "Types::Strict::String.constrained(format: /#{doc.pattern}/)"
```
mjgiarlo, line 111: This lets us reflect our OpenAPI regex patterns in Ruby models.

lib/cocina/generator/schema_value.rb
```diff
@@ -16,7 +16,7 @@ def generate
       private
 
       def type
-        "Types::#{dry_datatype(schema_doc)}#{optional}#{default}#{enum}"
+        "#{dry_datatype(schema_doc)}#{optional}#{default}#{enum}"
```
mjgiarlo, line 19: As above, the `Types::` prefixes have moved into `SchemaBase`.

lib/cocina/models/admin_policy.rb
```diff
@@ -11,10 +11,10 @@ class AdminPolicy < Struct
 
       # The version of Cocina with which this object conforms.
       # example: 1.2.3
-      attribute :cocinaVersion, Types::Strict::String.default(Cocina::Models::VERSION)
+      attribute :cocinaVersion, CocinaVersion.default(VERSION)
```
mjgiarlo, line 14: Case in point: instead of saying `cocinaVersion` should be a string, it should be a `CocinaVersion`. Note too that the explicit `Cocina::Models::` scope here was unnecessary since it's the default scope in this context.

lib/cocina/models/barcode.rb
```diff
@@ -0,0 +1,7 @@
+# frozen_string_literal: true
+
+module Cocina
+  module Models
+    Barcode = BusinessBarcode | LaneMedicalBarcode | CatkeyBarcode | StandardBarcode
+  end
+end
```
mjgiarlo, line 7: Here's what the new generated union types look like.

lib/cocina/models/catalog_link.rb
```diff
@@ -10,7 +10,7 @@ class CatalogLink < Struct
       attribute :refresh, Types::Strict::Bool.default(false)
       # Record identifier that is unique within the context of the linked record's catalog.
       # example: 11403803
-      attribute :catalogRecordId, Types::Strict::String
+      attribute :catalogRecordId, Types::Strict::String.constrained(format: /^\d+(:\d+)*$/)
```
mjgiarlo, line 13: Here you can see the regex lifted out of OpenAPI-land into Rubyland.

lib/cocina/generator/schema_array.rb
ndushay: 👌🏻

lib/cocina/generator/schema_base.rb
ndushay: bummer about the bug.  Thanks for the fix.

lib/cocina/generator/schema_base.rb
ndushay: excellent.

lib/cocina/generator/schema_value.rb
ndushay: 👌🏻

lib/cocina/models/barcode.rb
ndushay: schweet!

lib/cocina/models/admin_policy.rb
ndushay: reads more easily this way, for sure.

lib/cocina/models/catalog_link.rb
ndushay: nice.

lib/cocina/models/collection_identification.rb
```diff
@@ -7,7 +7,7 @@ class CollectionIdentification < Struct
       # Unique identifier in some other system. This is because a large proportion of what is deposited in SDR, historically and currently, are representations of objects that are also represented in other systems. For example, digitized paper and A/V collections have physical manifestations, and those physical objects are managed in systems that have their own identifiers. Similarly, books have barcodes, archival materials have collection numbers and physical locations, etc. The sourceId allows determining if an item has been deposited before and where to look for the original item if you're looking at its SDR representation. The format is: "namespace:identifier"
 
       # example: sul:PC0170_s3_Fiesta_Bowl_2012-01-02_210609_2026
-      attribute? :sourceId, Types::Strict::String
+      attribute? :sourceId, SourceId.optional
```
ndushay, line 10: can you explain the "optional" here?  what does it mean?  How was this expressed in `openapi.yml` to get here?  Why wasn't it "optional" before this change?

lib/cocina/models/identification.rb
```diff
@@ -4,14 +4,14 @@ module Cocina
...
+      attribute? :barcode, Barcode.optional
       attribute :catalogLinks, Types::Strict::Array.of(CatalogLink).default([].freeze)
       # Digital Object Identifier (https://www.doi.org)
-      attribute? :doi, Types::Nominal::Any
+      attribute? :doi, DOI.optional
       # Unique identifier in some other system. This is because a large proportion of what is deposited in SDR, historically and currently, are representations of objects that are also represented in other systems. For example, digitized paper and A/V collections have physical manifestations, and those physical objects are managed in systems that have their own identifiers. Similarly, books have barcodes, archival materials have collection numbers and physical locations, etc. The sourceId allows determining if an item has been deposited before and where to look for the original item if you're looking at its SDR representation. The format is: "namespace:identifier"
 
       # example: sul:PC0170_s3_Fiesta_Bowl_2012-01-02_210609_2026
-      attribute :sourceId, Types::Strict::String
+      attribute :sourceId, SourceId
```
ndushay, line 14: why is this `.optional` for collection identification but not here?  Because of openapi.yml?

ndushay (commented):
Thanks for the inline comments - very helpful.  Overall, seems like a very nice change.

I have some questions inline.

justinlittman (changes requested):
2 questions:
1. Have you verified that the openapi validator correctly handles union types?
2. Instead of adding more validation to the Ruby, shouldn't we just remove it all? It is redundant of the openapi validation (at best) and probably just slows everything down. (I admit that this question is a bit out of scope of this PR).

and a request: at least the most significant generator changes (e.g., union type) should have specs.

lib/cocina/models/collection_identification.rb
mjgiarlo: Ah, thanks for asking! I added this late yesterday after discovering a Cocina validation explosion on sdr-infra when testing this branch against stage. Here's what I saw there:

```shell
Validating Dro [▣▣▣▣▣▣▣▣▣▣▣▣▣▣▣▣▣▣▣⬚⬚⬚⬚⬚⬚⬚⬚⬚⬚⬚⬚] (61% (26000/42112), rate: 85.24/s, mean rate: 352.52/s,  1m13s total, ETA: 16:42:32)
/usr/local/rvm/gems/ruby-3.1.2/gems/dry-types-1.7.0/lib/dry/types/schema.rb:332:in `rescue in block in resolve_unsafe': Dry::Struct::
Error: [Cocina::Models::Identification.new] nil (NilClass) has invalid type for :barcode violates constraints (type?(String, nil) AND
 format?(/^36105[0-9]{9}$/, nil) failed) (Parallel::UndumpableException)
```

At the time, I interpreted this as, "I need to add code to allow barcodes to not only be missing but also to be nil." This is what the `.optional` method in dry-types does.

**But** maybe the problem here wasn't in the code and is instead anomalous stage data and I should try undoing this particular change and re-running against a bunch of objects in prod? (This change is a tiny part of the PR, so will be easy to remove.) If you don't have a hunch about this, who should we tag in to advise?

lib/cocina/models/identification.rb
mjgiarlo: Because of openapi.yml, yes. If a property of a component is not in the `required` array for the component, it's optional. Else, like here, it's required.

mjgiarlo:
@justinlittman 💬 
> 1. Have you verified that the openapi validator correctly handles union types?

Here's what I've done so far:

1. Validated the OpenAPI specification (via CI 👆🏻)
2. Run the specs successfully (via CI 👆🏻)
3. Ran `bin/validate-cocina` against this branch on `sdr-infra`

Is there another way to test in the way you're asking about? If so, happy to do it.

> 2. Instead of adding more validation to the Ruby, shouldn't we just remove it all? It is redundant of the openapi validation (at best) and probably just slows everything down. (I admit that this question is a bit out of scope of this PR).

Agree that it's out of scope and that we should probably kick this around at some point.
 
> and a request: at least the most significant generator changes (e.g., union type) should have specs.

For sure. I'll admit that I opened the generator specs, didn't grok how they were testing what they purported to be testing, and shrugged. I'll give it a deeper look today.

lib/cocina/models/collection_identification.rb
mjgiarlo: I can no longer reproduce this so I yanked the change around generating optional custom types.

justinlittman (approved):

PR #571, opened 2023-02-15
Allow non-required custom types to be optional

mjgiarlo:

Custom types in Ruby models used to be "any" types, which allow null values. This change allows e.g. barcodes to be nulled out or reset. I discovered this deficiency in the models while rolling out cocina-models 0.87.0, and patched around it: https://github.com/sul-dlss/argo/blob/main/app/services/item_change_set_persister.rb#L134-L143. This cocina-models patch will allow the prior behavior of being able to reset these custom type values when the attribute is not required by adding logic around determining optionality.

- [X] CI 
- [X] `sdr-infra`, against full QA dataset
- [X] `sdr-infra`, against full stage dataset
- [X] `sdr-infra`, against 500K prod dataset

### DIFF
--- a/.rubocop.yml
+++ b/.rubocop.yml
@@ -392,3 +392,5 @@ RSpec/Rails/InferredSpecType: # new in 2.14
   Enabled: true
 RSpec/Rails/MinitestAssertions: # new in 2.17
   Enabled: true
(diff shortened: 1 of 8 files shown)


ndushay (approved):
approving but flagging @justinlittman as he had good comments on the previous PR

justinlittman (approved):

Issue #574, opened 2023-02-21
Cannot instantiate Folio catalog links from hash


justinlittman:
```
Dry::Struct::Error:
       [Cocina::Models::SymphonyCatalogLink.new] "folio" (String) has invalid type for :catalog violates constraints (included_in?(["symphony", "previous symphony"], "folio") failed)
```

justinlittman:
See test in https://github.com/sul-dlss/cocina-models/pull/575

cc @mjgiarlo

ndushay:
I think the problem is the missing alpha prefix for the folio identifier;  I am closing this ticket per slack conversation with Justin.  Please reopen if I'm wrong.

PR #575, opened 2023-02-21
Adds test for catalog links.

justinlittman:
refs #574

### DIFF
--- a/spec/cocina/models/catalog_link_spec.rb
+++ b/spec/cocina/models/catalog_link_spec.rb
@@ -1,0 +1,40 @@
+# frozen_string_literal: true
+
+require 'spec_helper'
+
+RSpec.describe 'Cocina::Models::CatalogLink' do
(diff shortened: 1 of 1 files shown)


spec/cocina/models/catalog_link_spec.rb
```diff
@@ -0,0 +1,40 @@
...
+  end
+
+  context 'with symphony and folio catalog links' do
+    let(:props) do
+      {
+        sourceId: 'sul:1234',
+        catalogLinks: [
+          {catalog: 'symphony', catalogRecordId: '12345', refresh: true},
+          {catalog: 'previous symphony', catalogRecordId: '34567', refresh: false},
+          {catalog: 'folio', catalogRecordId: '45678', refresh: false}
```
ndushay, line 31: a valid folio identifier would be `a45678` or `in45678` or `L45678`

ndushay (changes requested):

justinlittman:
User error.

PR #615, opened 2023-08-16
Adds support for lite (relaxed, unvalidated) models.

justinlittman:

Allow for only partially populated cocina models.

### DIFF
--- a/lib/cocina/generator/generator.rb
+++ b/lib/cocina/generator/generator.rb
@@ -24,6 +24,11 @@ module Cocina
           generate_for(schema) if schema
         end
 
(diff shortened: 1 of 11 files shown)


ndushay (commented):
should there be an openapi.yml change?

does this get back to implying that certain elements can be nil rather than [] ?

justinlittman:
This doesn't require an openapi change, as it does not change anything in the cocina data model.

In a lite model, the base attributes will be null when they are omitted. However, it doesn't change the general approach to `[]`.

ndushay (approved):

PR #623, opened 2023-09-18
use safe navigator for temporal encoding code mapping, and 2 small refactors

ndushay:
~~NOTE:  Changes to openapi.yml require updating openapi.yml for sdr-api and dor-services-app and generating models - see README.~~

See PR #sul-dlss/dor_indexing_app/pull/993 which is failing because specs have nils where cocina-models doesn't expect them for xform to MODS.

It adds a safe navigator so our fragments of cocina in specs can be translated without error into MODS.

This accommodates dor_indexing_app using stanford-mods code for some Argo Solr index fields (so they are the same as SearchWorks index fields).   Part of #sul-dlss/dor_indexing_app/pull/992

specs.

### DIFF
--- a/lib/cocina/models/mapping/to_mods/form.rb
+++ b/lib/cocina/models/mapping/to_mods/form.rb
@@ -9,14 +9,14 @@ module Cocina
           # NOTE: H2 is the first case of structured form values we're implementing
           H2_SOURCE_LABEL = 'Stanford self-deposit resource types'
           PHYSICAL_DESCRIPTION_TAG = {
-            'reformatting quality' => :reformattingQuality,
-            'form' => :form,
-            'media type' => :internetMediaType,
-            'extent' => :extent,
-            'digital origin' => :digitalOrigin,
-            'media' => :form,
             'carrier' => :form,
+            'digital origin' => :digitalOrigin,
+            'extent' => :extent,
+            'form' => :form,
             'material' => :form,
+            'media' => :form,
+            'media type' => :internetMediaType,
+            'reformatting quality' => :reformattingQuality,
             'technique' => :form
           }.freeze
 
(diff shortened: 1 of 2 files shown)


lib/cocina/models/mapping/to_mods/form.rb
```diff
@@ -183,7 +183,7 @@ def write_physical_description_form_values(form_values, form: nil)
               attributes = {
                 unit: unit_for(form_value)
               }.tap do |attrs|
-                if PHYSICAL_DESCRIPTION_TAG.fetch(form_type) == :form && form_type != 'form'
+                if PHYSICAL_DESCRIPTION_TAG.fetch(form_type, nil) == :form && form_type != 'form'
```
ndushay, line 186: this isn't fixing the problem.  I will need to invoke Arcadia.

lib/cocina/models/mapping/to_mods/title.rb
```diff
@@ -178,7 +178,7 @@ def write_structured(title:, title_info_attrs: {})
               title_parts_without_names.each do |title_part|
                 title_type = tag_name_for(title_part)
                 title_value = title_value_for(title_part, title_type, title)
-                xml.public_send(title_type, title_value) if title_part.note.blank?
+                xml.public_send(title_type, title_value) if title_part.note.blank? && title_type
```
ndushay, line 181: this might be better addressed by putting a title in the pub_year spec in dor_indexing_app

lib/cocina/models/mapping/to_mods/subject.rb
```diff
@@ -298,7 +298,7 @@ def geographic(subject, subject_value, is_parallel: false)
 
           def time_range(subject)
             subject.structuredValue.each do |point|
-              xml.temporal point.value, point: point.type, encoding: subject.encoding.code
+              xml.temporal point.value, point: point.type, encoding: subject.encoding&.code
```
ndushay, line 301: this will do no harm, and helps the dor_indexing_app PR

lib/cocina/models/mapping/to_mods/form.rb
ndushay: (i did invoke Arcadia and we came up with a better solution in dor_indexing_app)

lib/cocina/models/mapping/to_mods/title.rb
ndushay: we fixed the problem in the dor_indexing_app specs.

justinlittman (approved):

PR #626, opened 2023-09-20
If an event note has a parallelValue, extract the data

aaron-collier:

While working on https://github.com/sul-dlss/dor_indexing_app/issues/992 for resource_type - I discovered that while we follow the same logic internally to determine this should provide a resource_type of `Journal/Periodical`, when translating the cocina to mods in order to use `stanford-mods` to determine the resource_type we do not look for the parallelValue and thus miss `value="monthly"` and `type="frequency"`.

This checks if the note has a parallelValue and if so, uses value(s) found there to write the note.

The Cocina for a note that has a parallelValue:

```
#<Cocina::Models::DescriptiveValue 
    structuredValue=[]
    parallelValue=[
        #<Cocina::Models::DescriptiveValue 
            structuredValue=[] parallelValue=[] groupedValue=[] 
            value="monthly" 
            type="frequency" 
            status=nil code=nil uri=nil standard=nil encoding=nil 
            identifier=[] source=nil displayLabel=nil qualifier=nil note=[] valueLanguage=nil valueAt=nil appliesTo=[]>
```

unit

### DIFF
--- a/lib/cocina/models/mapping/to_mods/event.rb
+++ b/lib/cocina/models/mapping/to_mods/event.rb
@@ -87,9 +87,15 @@ module Cocina
           # rubocop:enable Metrics/ParameterLists
 
           def write_note(note)
-            attributes = {}
-            attributes[:authority] = note.source.code if note&.source&.code
-            xml.send(note_tag_for(note), note.value, attributes)
+            if note.parallelValue.present?
+              note.parallelValue.each do |parallel_note|
+                write_note(parallel_note)
+              end
+            else
+              attributes = {}
+              attributes[:authority] = note.source.code if note&.source&.code
+              xml.send(note_tag_for(note), note.value, attributes)
+            end
           end
 
           def note_tag_for(note)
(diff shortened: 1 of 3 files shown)


lib/cocina/models/mapping/to_mods/event.rb
```diff
@@ -80,12 +80,22 @@ def write_event(cocina_event_type, dates, locations, names, notes, attributes, i
                 xml.publisher name.value, name_attributes(name) if name
               end
               Array(notes).each do |note|
-                write_note(note)
+                if note.parallelValue.present?
+                  write_parallel_note(note)
```
justinlittman, line 84: Wondering why this is showing up as untested?

spec/cocina/models/mapping/to_mods/note_spec.rb
```diff
@@ -42,4 +43,23 @@
       XML
     end
   end
+
+  context 'when notes has a parallelValue' do
+    let(:notes) do
+      [
+        Cocina::Models::DescriptiveValue.new(parallelValue: [Cocina::Models::DescriptiveValue.new(value: 'monthly', type: 'frequency')])
```
justinlittman, line 50: Shouldn't a parallelValue have at least 2 values?

lib/cocina/models/mapping/to_mods/event.rb
aaron-collier: We don't have a test for event.rb here at all... trying to simplify this logic just a bit to avoid the coverage issue.

spec/cocina/models/mapping/to_mods/note_spec.rb
aaron-collier: Maybe, in the case I'm running into there is only one.

arcadiafalcone (approved):

mjgiarlo (approved):

PR #633, opened 2023-10-16
Fix description types validation

justinlittman:

Types validation was not being performed.

Unit, qa, stage, and prod.


ndushay (commented):
should we have a test for  "type" and ":type" ?  do we already?

justinlittman:
`type` and `:type` and everything are normalized to string in `Cocina::Models::Validators::Validator` with call to `.with_indifferent_access`.

PR #639, opened 2023-11-01
Validate if values are whitespace only

aaron-collier:

This is draft while wrapping up styling/refactoring...

### DIFF
--- a/lib/cocina/models/validators/description_values_validator.rb
+++ b/lib/cocina/models/validators/description_values_validator.rb
@@ -53,9 +53,8 @@ module Cocina
         end
 
         def validate_values_for_blanks(hash, path)
-          return unless hash[:value] && hash[:value].is_a?(String) && hash[:value].blank? # rubocop:disable Style/SafeNavigation
-          puts hash
-          puts "hash[:value] = #{hash[:value]}"
+          return unless hash[:value] && hash[:value].is_a?(String) && /\A\s+\z/.match?(hash[:value]) # rubocop:disable Style/SafeNavigation
+
           error_paths_blank << path_to_s(path)
         end
 
--- a/lib/cocina/models/validators/description_values_validator.rb
+++ b/lib/cocina/models/validators/description_values_validator.rb
@@ -12,7 +12,8 @@ module Cocina
         def initialize(clazz, attributes)
           @clazz = clazz
           @attributes = attributes
-          @error_paths = []
+          @error_paths_multiple = []
+          @error_paths_blank = []
         end
 
         def validate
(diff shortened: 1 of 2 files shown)


lib/cocina/models/validators/description_values_validator.rb
```diff
@@ -12,29 +12,32 @@ def self.validate(clazz, attributes)
...
         def validate
           return unless meets_preconditions?
 
           validate_obj(attributes, [])
 
-          return if error_paths.empty?
+          # return if error_paths_multiple.empty? && error_paths_blank.empty?
 
-          raise ValidationError, "Multiple value, groupedValue, structuredValue, and parallelValue in description: #{error_paths.join(', ')}"
+          raise ValidationError, "Multiple value, groupedValue, structuredValue, and parallelValue in description: #{error_paths_multiple.join(', ')}" unless error_paths_multiple.empty?
```
jcoyne, line 26: Should it be able to return both kinds of errors at once?

lib/cocina/models/validators/description_values_validator.rb
```diff
@@ -12,29 +12,32 @@ def self.validate(clazz, attributes)
...
+          raise ValidationError, "Blank value, groupedValue, structuredValue, and parallelValue in description: #{error_paths_blank.join(', ')}" unless error_paths_blank.empty?
```
jcoyne, line 27:
```suggestion
          raise ValidationError, "Blank value in description: #{error_paths_blank.join(', ')}" unless error_paths_blank.empty?
```

lib/cocina/models/validators/description_values_validator.rb
aaron-collier: I don't think that's possible, at least as written. The multiples requires the top level `value` to be present and it's not when blank. Perhaps this is a topic for follow up discussion with @andrewjbtw or @arcadiafalcone

mjgiarlo:
Wondering if this is the sort of change we might want to run against a sizable subset of prod data to see how this affects validation.

justinlittman:
This should be validated against all items in all environments else chaos ensues.

aaron-collier:
@mjgiarlo / @justinlittman agreed. I've marked as hold.

jcoyne:
@aaron-collier how long will this take to validate?  We're getting hammered with these errors again/still.

lib/cocina/models/validators/description_values_validator.rb
```diff
@@ -51,10 +52,16 @@ def validate_obj(obj, path)
           validate_array(obj, path) if obj.is_a?(Array)
         end
 
-        def validate_values(hash, path)
+        def validate_values_for_blanks(hash, path)
+          return unless hash[:value] && hash[:value].is_a?(String) && hash[:value].blank? && !hash[:value].empty? # rubocop:disable Style/SafeNavigation
```
aaron-collier, line 56: @jcoyne I had to switch from your suggested regex to this `.blank? && !.empty?` because it was catching values that had consecutive newline characters (thus a begin-line/new-line/end-line match happens within a non blank string. If you have a better suggestion than what I have here, let me know.

jcoyne: I think I had anchored my regex with `^` and `$`, but I should have used `\A` and `\z` instead the former are end of line and the latter are end of string.

lib/cocina/models/validators/description_values_validator.rb
```diff
@@ -51,10 +52,16 @@ def validate_obj(obj, path)
           validate_array(obj, path) if obj.is_a?(Array)
         end
 
-        def validate_values(hash, path)
+        def validate_values_for_blanks(hash, path)
+          return unless hash[:value] && hash[:value].is_a?(String) && /\A\s+$\z/.match?(hash[:value]) # rubocop:disable Style/SafeNavigation
```
jcoyne, line 56: I think the $ can be removed in the regex

aaron-collier:
@andrewjbtw here is a list of the druids that vail cocina validation with this change. 1073 objects. Should these be remediated before this change goes in?

[validate-cocina.csv](https://github.com/sul-dlss/cocina-models/files/13261355/validate-cocina.csv)

aaron-collier:
Removing hold per @andrewjbtw:

```
I think these are all fixed. I’m not sure what will happen if MARC-origin updates trigger this issue after we turn on the validation, but I don’t see how we can chase down where those past errors came from. It’s worth turning on the validation for spreadsheet imports and then seeing how we can catch the MARC-related issues if they happen again.
```

@mjgiarlo / @justinlittman ready for a follow up review.

justinlittman (approved):

PR #640, opened 2023-11-01
add BCP 47 language tag attribute for files

### LINKED ISSUE #637: Add BCP 47 language field to the file model.
We need this to disambiguate multiple language transcriptions for the Nuremberg Trial Archive.  This data will be used to drive the options on the closed caption select widget for the media viewer.

https://www.rfc-editor.org/rfc/rfc4646.txt

jmartin-sul:
* optional because this won't be applicable for most files
* validated by the i18n gem if present
* initially useful for specifying the language of e.g. caption and transcript files

closes #637

unit tests


.rubocop.yml
```diff
@@ -116,6 +116,9 @@ Naming/PredicateName:
 
 # ----- RSpec ------
 
+RSpec/NestedGroups:
+  Max: 5
```
jmartin-sul, line 120: rubocop complained about too much example nesting in the spec i added (4 levels where the default limit is 3), and there were already a couple exceptions to this rule in `.rubocop_todo.yml`, so i just bumped this to what felt like a reasonable default and re-generated the todo file.

lib/cocina/models/validators/language_tag_validator.rb
```diff
@@ -0,0 +1,50 @@
...
+        rescue NameError
+          false
+        end
+
+        def valid_language_tag?
+          parsed_tag = I18n::Locale::Tag::Rfc4646.tag(attributes[:languageTag])
+
+          return true if parsed_tag.present? && parsed_tag.is_a?(I18n::Locale::Tag::Rfc4646)
+
+          false
```
mjgiarlo, line 45: This appears to be logically equivalent to:

```suggestion
          parsed_tag.present? && parsed_tag.is_a?(I18n::Locale::Tag::Rfc4646)
```

but maybe the explicit boolean returns communicate the intention more clearly?

mjgiarlo (approved):
Left a comment I'm on the fence about. Feel free to consider or not! Approving either way.

.rubocop_todo.yml
```diff
@@ -57,18 +57,11 @@ Metrics/ParameterLists:
 RSpec/DescribeClass:
   Enabled: false
 
-# Offense count: 87
+# Offense count: 224
```
jmartin-sul, line 60: i'm happy to bump this to a higher and more reasonable default if people want, either in a commit in this PR or in a separate PR.

this update came because i re-generated this TODO file after adding the setting override i made in `.rubocop.yml` -- it doesn't have anything to do directly with my changes in this PR.

the only reason i haven't already added an override for this rule in this PR is that i know some of us find changes like that to be extra noise when reviewing.

.rubocop_todo.yml
```diff
@@ -57,18 +57,11 @@ Metrics/ParameterLists:
...
 # Configuration parameters: CountAsOne.
 RSpec/ExampleLength:
   Max: 103
 
-# Offense count: 10
-# Configuration parameters: Max, AllowedGroups.
-RSpec/NestedGroups:
-  Exclude:
-    - 'spec/cocina/models/mapping/normalizers/mods/origin_info_normalizer_spec.rb'
-    - 'spec/cocina/models/validators/date_time_validator_spec.rb'
```
jmartin-sul, line 70: obviated by the `.rubocop.yml` change

.rubocop_todo.yml
```diff
@@ -99,7 +92,7 @@ Style/MultilineBlockChain:
     - 'lib/cocina/models/mapping/to_mods/form.rb'
     - 'lib/cocina/models/mapping/to_mods/subject.rb'
 
-# Offense count: 239
+# Offense count: 249
```
jmartin-sul, line 95: some long lines of less than 267 chars were added since the last `.rubocop_todo.yml` generation? 🤷

lib/cocina/models/file.rb
```diff
@@ -22,6 +22,8 @@ class File < Struct
       attribute :version, Types::Strict::Integer
       # MIME Type of the File.
       attribute? :hasMimeType, Types::Strict::String
+      # BCP 47 language tag: https://www.rfc-editor.org/rfc/rfc4646.txt -- other applications (like media players) expect language codes of this format, see e.g. https://videojs.com/guides/text-tracks/#srclang
+      attribute? :languageTag, Types::Strict::String.optional
```
jmartin-sul, line 26: i wasn't sure whether `.optional` was redundant given `attribute?`, though i see `presentation` also uses both of those.

fwiw, i noticed the `Changes to openapi.yml require updating openapi.yml for sdr-api and dor-services-app and generating models - see README` admonishment in the PR template, ran `exe/generator generate`, and this line dropped the `.optional`, turning into `attribute? :languageTag, Types::Strict::String`.  and the tests (which check that the field is optional) still passed. 🤷 (running that was also how i noticed that the comment here comes from the `openapi.yml` param `description`, so i lined those up to keep the diff between `exe/generator generate` and this PR to a minimum).

spec/cocina/models/validators/language_tag_validator_spec.rb
```diff
@@ -0,0 +1,117 @@
...
+      administrative: {
+        publish: false,
+        shelve: false,
+        sdrPreserve: true
+      },
+      hasMessageDigests: [],
+      hasMimeType: 'text/plain',
+      filename: 'page1.txt'
+    }
+  end
```
jmartin-sul, line 26: basically lifted from `dark_validator_spec.rb` to start, and collapsed some of its `let` vars that wouldn't change in this spec into the base properties.  didn't try to simplify more than that, but happy to do so if that's possible/preferable.

jmartin-sul (commented):
one other uncommitted diff between this PR and the cumulative changes that result when i run `exe/generator generate`: 

```diff
diff --git a/docs/description_types.md b/docs/description_types.md
index 5cd77d4..5e7723c 100644
--- a/docs/description_types.md
+++ b/docs/description_types.md
@@ -363,6 +363,8 @@ _Path: identifier.type_
   * document number
   * DOI
   * druid
+  * FOLIO
+    * FOLIO HRID for the source record of the metadata.
   * GTIN-14 ID
   * Handle
   * inventory number
```

i'm not sure if that indicates that a `exe/generator generate` run was missed for a prior change?  it wasn't clear to me from the README whether we always want to run `exe/generator generate`, or whether making all the changes manually can be sufficient.  but i figured good to surface this, because it's probably good to have less deviation between the `exe/generator generate` and the current state of this repo?

lib/cocina/models/validators/language_tag_validator.rb
jmartin-sul: ah, no, i like your suggestion, will take that.  i mean, what's there is _maybe_ very slightly clearer, but it was an artifact of iterating on this work and not something i intentionally did.  and being concise and relying on the implicit return seems more idiomatic for ruby.

jmartin-sul: fixed up the existing commit with this change

.rubocop.yml
mjgiarlo: 👍🏻

jmartin-sul:
merging since @mjgiarlo approved and i took his non-blocking suggestion.

not urgent, but still interested in feedback for possible follow on touchups re:
- `docs/description_types.md` after running `exe/generator generate`
- whether `attribute?` makes `optional` redundant
- whether to up the limit on (or disable) `RSpec/ExampleLength`

mjgiarlo:
@jmartin-sul 💬 
> merging since @mjgiarlo approved and i took his non-blocking suggestion.

🙏🏻 

> * `docs/description_types.md` after running `exe/generator generate`

I think we ought to take whatever the generator generates (now and always---and if the generator isn't generating the right stuff, we should fix the generator logic).

> * whether `attribute?` makes `optional` redundant

I believe `attribute?` makes the key optional (can be missing) and `optional` makes a value `nil`able?

> * whether to up the limit on (or disable) `RSpec/ExampleLength`

I'm 💯 on-board with the change you made.

jmartin-sul:
> @jmartin-sul 💬
> 
> > merging since @mjgiarlo approved and i took his non-blocking suggestion.
> 
> 🙏🏻
> 
> > * `docs/description_types.md` after running `exe/generator generate`
> 
> I think we ought to take whatever the generator generates (now and always---and if the generator isn't generating the right stuff, we should fix the generator logic).
> 

ok, took that output and added it in the first commit in https://github.com/sul-dlss/cocina-models/pull/641

made the `exe/generator generate` based change for the new model attributes too, but interestingly...

> > * whether `attribute?` makes `optional` redundant
> 
> I believe `attribute?` makes the key optional (can be missing) and `optional` makes a value `nil`able?
> 

mildly interesting find: to test this supposition out, i added the following to `language_tag_validator_spec.rb`

```ruby
  context 'with an empty for languageTag specified' do
    let(:props) do
      file_props.dup.tap do |props|
        # props[:languageTag] = ' '
        # props[:languageTag] = ''
        props[:languageTag] = nil
      end
    end

    context 'with a File' do
      let(:clazz) { Cocina::Models::File }

      it 'raises an error' do
        # expect(props[:languageTag]).to eq ' '
        # expect(props[:languageTag]).to eq ''
        expect(props[:languageTag]).to be_nil
        expect { validate }.to raise_error(Exception)
      end
    end

    context 'with a RequestFile' do
      let(:clazz) { Cocina::Models::RequestFile }

      it 'raises an error' do
        # expect(props[:languageTag]).to eq ' '
        # expect(props[:languageTag]).to eq ''
        expect(props[:languageTag]).to be_nil
        expect { validate }.to raise_error(Exception)
      end
    end
  end
```

and the test failed with `Expected block to raise error #<Exception>` for both of the test cases.  as the commented lines imply, i also tried `' '` and empty string in addition to `nil`, to see if any values that `activesupport` would treat as blank would get rejected by the model, but all had the same behavior of getting accepted as values.  am i writing this test case incorrectly?

i don't know if i should add a guard against blank/empty values in `LanguageTagValidator`?

at any rate, the tests already ensure that the value can be omitted entirely, so that does seem to indicate that `.optional` is superfluous in some way in this case.

> > * whether to up the limit on (or disable) `RSpec/ExampleLength`
> 
> I'm 💯 on-board with the change you made.

PR #641, opened 2023-11-02
touchups from file languageTag PR discussion

jmartin-sul:

make the couple agreed upon touchups from https://github.com/sul-dlss/cocina-models/pull/640#issuecomment-1790929840

existing unit tests (`language_tag_validator_spec.rb` already confirms that `languageTag` can be omitted from `Cocina::Models::File` and `Cocina::Models::RequestFile` properties).

### DIFF
--- a/lib/cocina/models/file.rb
+++ b/lib/cocina/models/file.rb
@@ -23,7 +23,7 @@ module Cocina
       # MIME Type of the File.
       attribute? :hasMimeType, Types::Strict::String
       # BCP 47 language tag: https://www.rfc-editor.org/rfc/rfc4646.txt -- other applications (like media players) expect language codes of this format, see e.g. https://videojs.com/guides/text-tracks/#srclang
-      attribute? :languageTag, Types::Strict::String.optional
+      attribute? :languageTag, Types::Strict::String
       # Use for the File.
       attribute? :use, Types::Strict::String
       attribute :hasMessageDigests, Types::Strict::Array.of(MessageDigest).default([].freeze)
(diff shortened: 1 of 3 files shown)


lib/cocina/models/file.rb
```diff
@@ -23,7 +23,7 @@ class File < Struct
       # MIME Type of the File.
       attribute? :hasMimeType, Types::Strict::String
       # BCP 47 language tag: https://www.rfc-editor.org/rfc/rfc4646.txt -- other applications (like media players) expect language codes of this format, see e.g. https://videojs.com/guides/text-tracks/#srclang
-      attribute? :languageTag, Types::Strict::String.optional
+      attribute? :languageTag, Types::Strict::String
```
jmartin-sul, line 26: probably not necessary to evaluating this PR, since there's a test that covers the omit-ability of this attribute.  but i was curious about the nuances of whether `attribute?` and `.optional` are redundant, and @mjgiarlo had a hypothesis about that, which i wrote a quick test for, but either i misunderstood and wrote the test wrong, or the difference if any wasn't quite what was suggested.  test and commentary in this comment on the prior PR: https://github.com/sul-dlss/cocina-models/pull/640#issuecomment-1791501666

mjgiarlo: @jmartin-sul Did the `.optional` come from the generator or did you patch that in earlier? I'm wondering if this suggests we could make a generator tweak based on what you've learned here.

jmartin-sul: i put the `.optional` in as part of my original manual implementation, and the generator took it out when i ran it after that (after checking the README to refresh myself on the right way to make model changes, which i probably should've done _before_ i started implementing 🙂)

mjgiarlo: OK, ha, excellent.

mjgiarlo (approved):

PR #643, opened 2023-11-08
switch LanguageTagValidator to operate on DRO::TYPES so that it's invoked automatically when building item models

jmartin-sul:

this makes `LanguageTagValidator` operate more like `DarkValidator`, which we determined in [slack discussion](https://stanfordlib.slack.com/archives/C09M7P91R/p1699289637148319) is likely desirable for real world validation usage:

@mjgiarlo 💬 
> What I think is happening: only some models use the ruby validation code (items, collections, APOs), and I don't think files are in that list. So they're strictly openapi validated.
>
> We likely need to rewrite the language tag validator so that it works against items instead of against files (so it runs at the item level and descends down into files, just like the dark-access validator)
> ...
> (and we can't lean on openapi validation for BCP unless we want to couple a given cocina-models release to a  an enum of allowed BCP values, which is maybe a bit brittle)
> ...
> It's a pretty small code change for a cocina-models patch, I think
> to rework the language tag validator to operate against DROs like the https://github.com/sul-dlss/cocina-models/blob/35ef3bdf826f53e7fd6e158b2db3b85a8cfadd78/lib/cocina/models/validators/dark_validator.rb

@jmartin-sul 💬 
> ah, ok, that makes sense.  didn't realize validation would be invoked automatically on the argo side for only some model types, and there wasn't any overarching info needed at the item level (i don't think?) to validate the tag on an individual file, so i went for the simpler file level thing.

@mjgiarlo 💬 
> validation in cocina-models is maybe a bit surprising in this way. it's like that in the gem, too, not just argo. validation is wired up on the instance and class new methods: https://github.com/sul-dlss/cocina-models/blob/main/lib/cocina/models/validatable.rb
> but not every model uses the `Validatable` module. (`DRO` does. `File` doesn't.)

- [x] unit tests
- [ ] tweak the argo branch for [this PR](https://github.com/sul-dlss/argo/pull/4266) to use this cocina-models branch,  deploy to stage, and see if that fixes the lack of validation on `languageTag` mentioned in the linked slack discussion?  and if so, merge this PR and cut a new patch release and update the argo PR to use that?

### DIFF
--- a/lib/cocina/models/validators/language_tag_validator.rb
+++ b/lib/cocina/models/validators/language_tag_validator.rb
@@ -17,10 +17,10 @@ module Cocina
         def validate
           return unless meets_preconditions?
 
-          return if valid_language_tag?
+          return if invalid_files.empty?
 
-          raise ValidationError, 'The provided language tag is not valid according to RFC 4646: ' \
-                                 "#{attributes[:languageTag]}"
+          raise ValidationError, 'Some files have invalid language tags according to RFC 4646: ' \
+                                 "#{invalid_filenames_with_language_tags.join(', ')}"
         end
 
         private
@@ -28,19 +28,45 @@ module Cocina
         attr_reader :clazz, :attributes
 
         def meets_preconditions?
-          file? && attributes[:languageTag].present?
+          dro?
         end
 
-        def file?
-          (clazz::TYPES & File::TYPES).any?
+        def dro?
+          (clazz::TYPES & DRO::TYPES).any?
         rescue NameError
           false
         end
 
-        def valid_language_tag?
-          parsed_tag = I18n::Locale::Tag::Rfc4646.tag(attributes[:languageTag])
+        def valid_language_tag?(file)
+          # I18n::Locale::Tag::Rfc4646.tag will return an instance of I18n::Locale::Tag::Rfc4646 (with fields like language, script,
+          # region) for strings that can be parsed according to RFC 4646, and nil for strings that do not conform to the spec.
+          I18n::Locale::Tag::Rfc4646.tag(file[:languageTag]).present?
+        end
 
-          parsed_tag.present? && parsed_tag.is_a?(I18n::Locale::Tag::Rfc4646)
+        def invalid_files
+          @invalid_files ||= language_tag_files.reject { |file| valid_language_tag?(file) }
+        end
+
+        def invalid_filenames_with_language_tags
+          invalid_files.map { |invalid_file| "#{invalid_file[:filename] || invalid_file[:label]} (#{invalid_file[:languageTag]})" }
+        end
+
+        def language_tag_files
+          files.select { |file| file[:languageTag].present? }
+        end
+
+        def files
+          [].tap do |files|
+            next if attributes.dig(:structural, :contains).nil?
+
+            attributes[:structural][:contains].each do |fileset|
+              next if fileset.dig(:structural, :contains).nil?
+
+              fileset[:structural][:contains].each do |file|
+                files << file
+              end
+            end
+          end
         end
       end
     end
--- a/spec/cocina/models/validators/language_tag_validator_spec.rb
+++ b/spec/cocina/models/validators/language_tag_validator_spec.rb
@@ -3,7 +3,27 @@
 require 'spec_helper'
 
 RSpec.describe Cocina::Models::Validators::LanguageTagValidator do
-  let(:validate) { described_class.validate(clazz, props) }
+  let(:validate) { described_class.validate(clazz, dro_props) }
+
+  let(:dro_props) do
+    {
+      type: Cocina::Models::ObjectType.book,
+      access: { view: 'dark', download: 'none' },
+      structural: {
+        contains: [
+          {
+            externalIdentifier: 'bc123df4567_1',
+            label: 'Fileset 1',
+            type: Cocina::Models::FileSetType.file,
+            version: 1,
+            structural: {
+              contains: [props]
+            }
+          }
+        ]
+      }
+    }
+  end
 
   let(:props) { file_props }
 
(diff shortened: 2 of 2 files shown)


lib/cocina/models/validators/language_tag_validator.rb
```diff
@@ -17,31 +17,67 @@ def initialize(clazz, attributes)
...
+            next if attributes.dig(:structural, :contains).nil?
+
+            attributes[:structural][:contains].each do |fileset|
+              next if fileset.dig(:structural, :contains).nil?
+
+              fileset[:structural][:contains].each do |file|
+                files << file
+              end
+            end
+            puts "LanguageTagValidator#files: #{files}"
```
justinlittman, line 78: Cruft?

lib/cocina/models/validators/language_tag_validator.rb
```diff
@@ -17,31 +17,67 @@ def initialize(clazz, attributes)
...
+        def valid_language_tag?(file)
+          parsed_tag = I18n::Locale::Tag::Rfc4646.tag(file[:languageTag])
 
           parsed_tag.present? && parsed_tag.is_a?(I18n::Locale::Tag::Rfc4646)
         end
+
+        def invalid_files
+          @invalid_files ||=
+            [].tap do |invalid_files|
+              language_tag_files.each do |file|
```
justinlittman, line 49: Here and below. This seems more idiomatic than tap:
```
invalid_files = language_tag_files.reject {|file| valid_language_tag?(file) }
```

lib/cocina/models/validators/language_tag_validator.rb
```diff
@@ -17,31 +17,67 @@ def initialize(clazz, attributes)
...
+          invalid_files.map { |invalid_file| "#{invalid_file[:filename] || invalid_file[:label]} (#{invalid_file[:languageTag]})" }
+        end
+
+        def language_tag_files
+          [].tap do |language_tag_files|
+            files.each do |file|
+              language_tag_files << file if file[:languageTag].present?
+            end
+          end
+        end
```
mjgiarlo, line 65: Similar to above, wonder how this would stack up against:

```suggestion
        def language_tag_files
          files.select { |file| file[:languageTag].present? }
        end
 ```

File: lib/cocina/models/validators/language_tag_validator.rb [hunk 4]
```diff
@@ -17,31 +17,67 @@ def initialize(clazz, attributes)
...
         rescue NameError
           false
         end
 
-        def valid_language_tag?
-          parsed_tag = I18n::Locale::Tag::Rfc4646.tag(attributes[:languageTag])
+        def valid_language_tag?(file)
+          parsed_tag = I18n::Locale::Tag::Rfc4646.tag(file[:languageTag])
 
           parsed_tag.present? && parsed_tag.is_a?(I18n::Locale::Tag::Rfc4646)
```
mjgiarlo, line 43: Could the `is_a?` check suffice here? Presumably if the parsed tag is `nil` or an empty string, that will return `false`.

lib/cocina/models/validators/language_tag_validator.rb
jmartin-sul: whoops, yup!  will fix.

lib/cocina/models/validators/language_tag_validator.rb
jmartin-sul: ah yeah, i like that, i cribbed the current implementation off `DarkValidator`, and had meant to circle back to do something more concise/idiomatic, but obviously forgot some cleanup between this and the straggler `puts`.

lib/cocina/models/validators/language_tag_validator.rb
jmartin-sul: oh yeah, probably, will try...

jmartin-sul: that worked, also added a comment explaining expected return value from `I18n::Locale::Tag::Rfc4646.tag`

jmartin-sul:
@justinlittman and @mjgiarlo thanks for the feedback, implemented all suggestions, ready for another look

jmartin-sul:
fixed the method i ripped off to be more concise, using the suggestion from this PR: https://github.com/sul-dlss/cocina-models/pull/645

lib/cocina/models/validators/language_tag_validator.rb
```diff
@@ -17,30 +17,56 @@ def initialize(clazz, attributes)
...
         rescue NameError
           false
         end
 
-        def valid_language_tag?
-          parsed_tag = I18n::Locale::Tag::Rfc4646.tag(attributes[:languageTag])
+        def valid_language_tag?(file)
+          # I18n::Locale::Tag::Rfc4646.tag will return an instance of I18n::Locale::Tag::Rfc4646 (with fields like language, script,
+          # region) for strings that can be parsed according to RFC 4646, and nil for strings that do not conform to the spec.
+          I18n::Locale::Tag::Rfc4646.tag(file[:languageTag]).is_a?(I18n::Locale::Tag::Rfc4646)
```
justinlittman, line 43: Why not just `I18n::Locale::Tag::Rfc4646.tag(file[:languageTag])` as a nil check?

justinlittman (approved):

lib/cocina/models/validators/language_tag_validator.rb
jmartin-sul: oh, good question, no particular reason.  i think use of `.is_a?` is partially an artifact of playing with this on the console to see what the return type was for valid and invalid strings.

not having thought about it much, i may very slightly prefer this, because it's more explicit about the expected return value from the parser, but i don't feel strongly at all and am totally happy to just do a nil check if you'd prefer.

mjgiarlo: It seems like we care more about the return value of `I18n::Locale::Tag::Rfc4646#tag` being non-`nil` than being a legit `I18n::Locale::Tag::Rfc4646` struct instance (which we throw away). Do you think `#present?` would reflect better that this is what we care about?

mjgiarlo (approved):

lib/cocina/models/validators/language_tag_validator.rb
jmartin-sul: i think i still slightly prefer the `is_a?`, but since it's not a strong preference at all, i'm fine being outvoted, changed to `#present?` and re-pushed 🙂 (i do agree that we care more about non-nil here -- i think my pref mostly stems from the implicit guidance the `is_a?` provided if a cocina-models consumer wanted to parse the field, but the comment makes that same guidance more explicit, so the `is_a?` then feels even less necessary)

PR #644, opened 2023-11-08
Allow file use and language tag attrs to be removed

mjgiarlo:
# Why was this change made?

Prior to this commit, it was not possible to create a new instance of a `File` model containing use or language tag attrs with either or both removed (replaced by `nil`s). See https://github.com/sul-dlss/argo/blob/main/app/services/structure_updater.rb#L92, which prevents users from removing file use values. 

This commit makes it possible to do so.

Because this makes our models slightly more permissive, I don't think running a validation in a deployed env is worth the time/cost.

# How was this change tested?

CI, `bundle console`


justinlittman (approved):

ndushay:
is this anything andrew or arcadia would care about?  (I doubt it?)

jmartin-sul (approved):

mjgiarlo:
@ndushay 💬 
> is this anything andrew or arcadia would care about? (I doubt it?)

Andrew wants to be able to unset file use and language attrs and currently can't. This change only affects structural metadata, not descriptive.

PR #654, opened 2023-12-19
add methods for extracting "main title" from cocina descriptive

### LINKED ISSUE #653: method to get main_title
part of sul-dlss/dor_indexing_app/issues/1075

We want to be able to boost relevancy in Argo if there is an exact match on the short main title, akin to the 245a boosting in SearchWorks relevancy:

from sul-dlss/sul-solr-configs/blob/master/searchworks-prod/solrconfig.xml:

```xml
      <str name="qf">
        title_245a_exact_search^1000
        title_245a_unstem_search^500
        title_245a_search^75           vern_title_245a_search^75
        title_245_unstem_search^75
        title_245_search^50            vern_title_245_search^50
```

To distinguish between the short main title, 245a, and the full main title, I got this spec from Arcadia:

"245a = title.structuredValue.value with type 'main title' (or just title.value), 245b = title.structuredValue.value with type 'subtitle'; if there are multiple titles, the 245 should have status 'primary' or at least be first"

ndushay:

To support nuanced indexing of titles for Argo, per sul-dlss/dor_indexing_app/issues/1044

Closes #653

CI.  

This adds a method to be used by dor_indexing app, with a direct call to the cocina method.

### DIFF
--- a/lib/cocina/models/builders/title_builder.rb
+++ b/lib/cocina/models/builders/title_builder.rb
@@ -24,22 +24,39 @@ module Cocina
           new(strategy: strategy, add_punctuation: add_punctuation).build(titles)
         end
 
+        # the "main title" is the title withOUT subtitle, part name, etc.  We want to index it separately so
+        #   we can boost matches on it in search results (boost matching this string higher than matching full title string)
+        #   e.g. "The Hobbit" (main_title) vs "The Hobbit, or, There and Back Again (full_title)
+        # @param [[Array<Cocina::Models::Title,Cocina::Models::DescriptiveValue>] titles the titles to consider
+        # @return [String] the main title value for Solr
+        def self.main_title(titles)
+          new(strategy: :first, add_punctuation: false).main_title(titles)
+        end
+
         def initialize(strategy:, add_punctuation:)
           @strategy = strategy
           @add_punctuation = add_punctuation
         end
 
-        # @param [[Array<Cocina::Models::Title>] titles the titles to consider
+        # @param [[Array<Cocina::Models::Title>] cocina_titles the titles to consider
         # @return [String] the title value for Solr
-        def build(titles)
-          cocina_title = primary_title(titles) || untyped_title(titles)
-          cocina_title = other_title(titles) if cocina_title.blank?
+        def build(cocina_titles)
+          cocina_title = primary_title(cocina_titles) || untyped_title(cocina_titles)
+          cocina_title = other_title(cocina_titles) if cocina_title.blank?
 
           if strategy == :first
             extract_title(cocina_title)
           else
-            cocina_title.map { |one| extract_title(one) }
+            cocina_titles.map { |one| extract_title(one) }
           end
+        end
+
+        def main_title(titles)
+          cocina_title = primary_title(titles) || untyped_title(titles)
+          cocina_title = other_title(titles) if cocina_title.blank?
+
+          cocina_title = cocina_title.first if cocina_title.is_a?(Array)
+          extract_main_title(cocina_title)
         end
 
         private
@@ -55,6 +72,16 @@ module Cocina
                      return build(cocina_title.parallelValue)
                    end
           remove_trailing_punctuation(result.strip) if result.present?
+        end
+
+        def extract_main_title(cocina_title)
+          if cocina_title.value
+            cocina_title.value # covers both title and main title types
+          elsif cocina_title.structuredValue.present?
+            main_title_from_structured_values(cocina_title)
+          elsif cocina_title.parallelValue.present?
+            main_title(cocina_title.parallelValue)
+          end
         end
 
         def add_punctuation?
@@ -102,15 +129,16 @@ module Cocina
           end
         end
 
-        # rubocop:disable Metrics/BlockLength
-        # rubocop:disable Metrics/CyclomaticComplexity
-        # rubocop:disable Metrics/PerceivedComplexity
-        # rubocop:disable Metrics/MethodLength
         # @param [Cocina::Models::Title] title with structured values
         # @return [String] the title value from combining the pieces of the structured_values by type and order
         #   with desired punctuation per specs
+        #
+        # rubocop:disable Metrics/BlockLength
+        # rubocop:disable Metrics/CyclomaticComplexity
+        # rubocop:disable Metrics/MethodLength
+        # rubocop:disable Metrics/PerceivedComplexity
         def title_from_structured_values(title)
-          structured_title = ''
+          result = ''
           part_name_number = ''
           # combine pieces of the cocina structuredValue into a single title
           title.structuredValue.each do |structured_value|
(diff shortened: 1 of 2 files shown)


lib/cocina/models/builders/title_builder.rb
```diff
@@ -102,15 +135,16 @@ def other_title(titles)
...
-        # rubocop:disable Metrics/PerceivedComplexity
-        # rubocop:disable Metrics/MethodLength
         # @param [Cocina::Models::Title] title with structured values
         # @return [String] the title value from combining the pieces of the structured_values by type and order
         #   with desired punctuation per specs
+        #
+        # rubocop:disable Metrics/BlockLength
+        # rubocop:disable Metrics/CyclomaticComplexity
+        # rubocop:disable Metrics/MethodLength
+        # rubocop:disable Metrics/PerceivedComplexity
```
ndushay, line 145: moved these down and alphabetized

lib/cocina/models/builders/title_builder.rb
```diff
@@ -24,24 +24,46 @@ def self.build(titles, strategy: :first, add_punctuation: true)
...
+        #   e.g. "The Hobbit" (main_title) vs "The Hobbit, or, There and Back Again (full_title)
+        # @param [[Array<Cocina::Models::Title,Cocina::Models::DescriptiveValue>] titles the titles to consider
+        # @param [Symbol] strategy ":first" is the strategy for selection when primary or display
+        #   title are missing
+        # @param [Boolean] add_punctuation determines if the title should be formmated with punctuation
+        # @return [String] the title value for Solr
+        def self.main_title(titles)
+          # 245a = title.structuredValue.value with type 'main title' (or just title.value),
+          # 245b = title.structuredValue.value with type 'subtitle'; if there are multiple titles, the 245 should have status 'primary' or at least be first
+          new(strategy: :first, add_punctuation: false).main_title(titles)
```
mjgiarlo, line 38: Is the `:first` strategy used when invoking `#main_title`, or is it primarily used in `#build`?

lib/cocina/models/builders/title_builder.rb
```diff
@@ -24,24 +24,46 @@ def self.build(titles, strategy: :first, add_punctuation: true)
...
-          cocina_title = other_title(titles) if cocina_title.blank?
+        def build(cocina_titles)
+          cocina_title = primary_title(cocina_titles) || untyped_title(cocina_titles)
+          cocina_title = other_title(cocina_titles) if cocina_title.blank?
 
           if strategy == :first
             extract_title(cocina_title)
           else
-            cocina_title.map { |one| extract_title(one) }
+            cocina_titles.map { |one| extract_title(one) }
```
mjgiarlo, line 55: Can you say more about this change? I think I grok the variable name changes in this method, but this (naïvely) appears to be a functional change.

lib/cocina/models/builders/title_builder.rb
```diff
@@ -57,6 +79,17 @@ def extract_title(cocina_title)
           remove_trailing_punctuation(result.strip) if result.present?
         end
 
+        def extract_main_title(cocina_title)
+          if cocina_title.value
+            # if cocina_title.type&.downcase == 'main title'
```
mjgiarlo, line 84: Useful or cruft?

lib/cocina/models/builders/title_builder.rb
```diff
@@ -125,40 +159,76 @@ def title_from_structured_values(title)
...
+            value = structured_value.value&.strip
+            next unless value
+
+            case structured_value.type&.downcase
+            when 'nonsorting characters'
+              non_sort_value = "#{value}#{non_sorting_padding(cocina_title, value)}"
+              result = if result.present?
+                         "#{non_sort_value}#{result}"
+                       else
+                         non_sort_value
```
mjgiarlo, line 221: If `result` is either `nil` or `''`, won't `"#{non_sort_value}#{result}"` be equivalent to `non_sort_value`? The condition check may be unnecessary.

lib/cocina/models/builders/title_builder.rb
```diff
@@ -125,40 +159,76 @@ def title_from_structured_values(title)
...
+                       end
+            when 'main title'
+              result = "#{result}#{value}"
+            when 'title'
+              result = value
+            end
+          end
+          result
```
mjgiarlo, line 229: I'm not sure we need the `result` variable here since the method will return the last evaluated value, which in each branch is the value of `result`.

lib/cocina/models/builders/title_builder.rb
mjgiarlo: In other words, I wonder if `:main` or `:main_title` could or should be a strategy all its own and be a branch within `#build` instead of its own new instance method.

ndushay: build will be changing in my next PR;  I'm not sure I want to optimize the code until I've got the correct strings for main_title, build, full_title (next PR) and additional_titles (next next PR)

lib/cocina/models/builders/title_builder.rb
ndushay: cruft - thanks.

lib/cocina/models/builders/title_builder.rb
ndushay: This method is broken, and this is one way.  Why would you call map on a single title object?

I did the rename for my own sanity and ... voila - it became clear that this should be the plural.

lib/cocina/models/builders/title_builder.rb
ndushay: good eye!  thanks.

lib/cocina/models/builders/title_builder.rb
ndushay: this is returning the value of result after the loop through all the pieces of the structured value.

ndushay (commented):
thanks for the review Mike - take another gander, please?

lib/cocina/models/builders/title_builder.rb
ndushay: Note that `build` is giving bad titles now in how it marries the pieces together for structured values.

mjgiarlo: Gotcha, thanks!

ndushay: to answer your specific question, "first" will be used by main_title and full_title but "all" will be used by additional title

lib/cocina/models/builders/title_builder.rb
ndushay: plus ... specs.

spec/cocina/models/builders/title_builder_spec.rb
```diff
@@ -30,7 +31,7 @@
       }
     end
 
-    it 'returns the first title object and logs a deprecation' do
+    it '.build returns the first title as a string and logs a deprecation' do
```
mjgiarlo, line 34: This change suggests that what's missing hereabouts is a `describe '.build' do` block containing these `it` tests. But since you are actively working through this class and its methods and have plans for follow-on work, I'm inclined to leave this feedback for later consideration, approve, and merge.

mjgiarlo (approved):

Issue #655, opened 2023-12-19
methods to get "full title" and "additional titles"


ndushay:
part of sul-dlss/dor_indexing_app/issues/1075 and sul-dlss/dor_indexing_app/issues/1044

We want to be able to boost relevancy in Argo distinguishing between matches on "main title", on "full title" (main title + subtitle and such), and on additional titles.   

Since the cocina gem already has a "TitleBuilder" with a lot of the logic present, it is easiest to add the new methods to this gem, rather than duplicate the logic elsewhere.  Also, it makes it available for other potential consumers.

ndushay:
Closing this, but there will be tweaks as an outgrowth of #657

PR #658, opened 2023-12-20
titleBuilder: additional_titles passes tests, a few more tweaks to .build method

ndushay:

To support nuanced indexing of titles for Argo, per https://github.com/sul-dlss/dor_indexing_app/issues/1044

part of https://github.com/sul-dlss/cocina-models/issues/655

CI.

This adds a method to be used by dor_indexing app, with a direct call to the cocina method.

### DIFF
--- a/spec/cocina/models/builders/title_builder_spec.rb
+++ b/spec/cocina/models/builders/title_builder_spec.rb
@@ -11,6 +11,7 @@ RSpec.describe Cocina::Models::Builders::TitleBuilder do
   let(:druid) { 'druid:bc753qt7345' }
   let(:main_title) { described_class.main_title(cocina_titles) }
   let(:full_title) { described_class.full_title(cocina_titles) }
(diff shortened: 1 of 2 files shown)


spec/cocina/models/builders/title_builder_spec.rb
```diff
@@ -161,14 +174,16 @@
...
+    end
   end
 
-  context 'when the titles are in a structuredValue' do
+  context 'when the title is in a structuredValue' do
     # from a spreadsheet upload integration test
     #   https://argo-stage.stanford.edu/view/sk561pf3505
     let(:titles) do
-      # Yes, there can be a structuredValue inside a StructuredValue.  For example,
-      # a uniform title where both the name and the title have internal StructuredValue
```
ndushay, line 171: this comment only applies in the one test where I left it;  this is copy pasta.

aaron-collier (approved):

PR #665, opened 2024-01-17
Update description_types.md

arcadiafalcone:

Additional types used in metadata


justinlittman (changes requested):
This file is automatically generated. The file that you probably want to update is description_types.yml.

PR #669, opened 2024-01-19
Enabled description types validator

justinlittman:

Validation is better than no validation.

This validation was checked against all production items.


edsu (approved):
Approving, assuming you are confident this won't set the SDR on fire.

PR #675, opened 2024-01-24
Validate for title.structuredValue.value having a type

lwrubel:
Validator to address https://github.com/sul-dlss/argo/issues/4279. This goes down one level of nesting (`title.structuredValue`) and also validates one level of `relatedResource.title.structuredValue`. 

Unit. Ran `bin/validate-cocina` on QA, stage, and prod and there were no existing objects with this problem. 

### DIFF
--- a/lib/cocina/models/validators/description_values_validator.rb
+++ b/lib/cocina/models/validators/description_values_validator.rb
@@ -68,7 +68,7 @@ module Cocina
         end
 
         def validate_title_type(hash, path)
-          # only apply to title.structuredValue or relatedResource.title.structuredValue with value
+          # only apply to title.structuredValue, title.parallelValue.structuredValue, or relatedResource.title with a value
           return unless hash[:value] && (path.first == :title || related_resource_title?(path)) && path.include?(:structuredValue)
 
           # if there is a "value" key, make sure there is also a "type" key, only for title.structuredValue
@@ -76,7 +76,8 @@ module Cocina
         end
 
         def related_resource_title?(path)
-          path.first == :relatedResource && path.third == :title
+          # title is directly within relatedResource, e.g [:relatedResource, 0, :title, 0, :structuredValue, 0])
+          path.first == :relatedResource && path[2] == :title
         end
 
         def path_to_s(path)
--- a/lib/cocina/models/validators/description_values_validator.rb
+++ b/lib/cocina/models/validators/description_values_validator.rb
@@ -14,6 +14,7 @@ module Cocina
           @attributes = attributes
           @error_paths_multiple = []
           @error_paths_blank = []
+          @error_paths_missing_title_type = []
         end
 
         def validate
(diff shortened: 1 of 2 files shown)


lib/cocina/models/validators/description_values_validator.rb
```diff
@@ -64,6 +67,14 @@ def validate_values_for_multiples(hash, path)
           error_paths_multiple << path_to_s(path)
         end
 
+        def validate_title_type(hash, path)
+          # only apply to title.structuredValue with value
+          return unless hash[:value] && path.include?(:title) && path.include?(:structuredValue)
```
justinlittman, line 72: Wondering if this might match too broadly. Can you check for "title" and "structuredValue" in particular locations in the path?

lwrubel: Thanks, I made some updates to this, also as part of checking `relatedResource` titles. There may be some ways cocina gets created that I'm not thinking of, however.

peetucket (approved):
LGTM

peetucket:
@justinlittman if you happen to have a chance before departing and wanted to double-check this work before merging

lwrubel:
@peetucket OK if I go ahead and merge this since I don't see any further notes from @justinlittman?

lib/cocina/models/validators/description_values_validator.rb
```diff
@@ -64,6 +67,18 @@ def validate_values_for_multiples(hash, path)
...
+        def validate_title_type(hash, path)
+          # only apply to title.structuredValue or relatedResource.title.structuredValue with value
+          return unless hash[:value] && (path.first == :title || related_resource_title?(path)) && path.include?(:structuredValue)
+
+          # if there is a "value" key, make sure there is also a "type" key, only for title.structuredValue
+          error_paths_missing_title_type << path_to_s(path) unless hash[:type]
+        end
+
+        def related_resource_title?(path)
+          path.first == :relatedResource && path.third == :title
```
ndushay, line 79: what is second?   structuredValue or parallelValue or something?  It's theoretically possible to have structuredValues inside parallelValues ...

ndushay (commented):
just a note about using "third".   ow. looks good.

lib/cocina/models/validators/description_values_validator.rb
peetucket: Is there an easy way to model that scenario in another spec and see what happens?  Do you think that should block this fix?

lwrubel: When I talked this through with Justin, we went down the path of keeping this in structuredValue only. Not further nesting or other situations. 

The path it's looking for would be `[:title, 0, :structuredValue, 0]` in order to validate the presence of a `type`. in the hash. 

Is there a way you'd recommend we do this? I don't doubt there's something I'm not understanding here.

lwrubel: @ndushay and I took a look at this and I'll add in the case to handle `title.parallelValue.structuredValue`. @arcadiafalcone says: this would likely be coming from a MARC record, so with Google Books it's not that uncommon.

lwrubel: @ndushay see added specs to cover `title.parallelValue.structuredValue` and `relatedResource.title.parallelValue.structuredValue` situations. Because the logic is to check for `structuredValue`  in the path, but not directly as a descendant of `title`, there is room for a `parallelValue` to exist in the path. I think this is OK for titles?

ndushay (approved):

PR #676, opened 2024-01-25
TitleBuilder gets detailed specs from dor_indexing gem

ndushay:

specs should live with the code;  I suspect TitleBuilder was moved from dor_indexing_app to cocina-models at some point and all the specs didn't come over with it.

I discovered these specs when I started to use the new "main_title" and "full_title" and "additional_titles" fields in dor_indexing gem.

CI.  Trivial code change;  it was just adding lots of specs.

### DIFF
--- a/lib/cocina/models/builders/title_builder.rb
+++ b/lib/cocina/models/builders/title_builder.rb
@@ -201,6 +201,16 @@ module Cocina
         # @return [String] the title value from combining the pieces of the structured_values by type and order
         #   with desired punctuation per specs
         #
+        # - nonsorting characters value is followed by a space, unless the nonsorting
+        #   character count note has a numeric value equal to the length of the
+        #   nonsorting characters value, in which case no space is inserted
+        # - subtitle is preceded by space colon space, unless it is at the beginning
+        #   of the title string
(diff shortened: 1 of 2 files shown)


lib/cocina/models/builders/title_builder.rb
```diff
@@ -245,7 +255,7 @@ def rebuild_structured_value(cocina_title)
                        end
             when 'subtitle'
               result = if !add_punctuation?
-                         [result, value].select(&:presence).join(' ')
+                         [result, value.sub(/^:/, '').strip].select(&:presence).join(' ')
```
ndushay, line 258: addresses one spec I copied over;  my new title variants required this tweak;  the old method for getting a display title was unaffected.

ndushay (commented):
Note:  no existing specs were removed from title_builder_spec except for one duplicate (which I probably introduced in earlier work).  Some were moved to a different spot in the file;  some were tweaked.

I tried to organize the specs a little better and I tried to be consistent, up to a point.

In some cases I changed the title value strings because my eyes started swimming with "title" .

At this point, I am thoroughly sick of these specs, but they will be a great help if someone needs to tweak our title building code.

peetucket (approved):

PR #681, opened 2024-02-05
[DRAFT] Fix validation of description title

lwrubel:
Follow-on to #675 to make sure that ValidationError is raised in several invalid title situations. 

Unit. When running a dor-services-app instance locally in docker with the cocina-models branch, error is raised when sent from Argo spreadsheet upload. Currently running `bin/validate-cocina`.

### DIFF
--- a/lib/cocina/models/validators/description_values_validator.rb
+++ b/lib/cocina/models/validators/description_values_validator.rb
@@ -62,22 +62,30 @@ module Cocina
         end
 
         def validate_values_for_multiples(hash, path)
-          return unless hash.count { |key, value| %i[value groupedValue structuredValue parallelValue].include?(key) && value.present? } > 1
+          return unless hash.count { |key, value| %w[value groupedValue structuredValue parallelValue].include?(key) && value.present? } > 1
 
           error_paths_multiple << path_to_s(path)
         end
 
         def validate_title_type(hash, path)
           # only apply to title.structuredValue, title.parallelValue.structuredValue, or relatedResource.title with a value
-          return unless hash[:value] && (path.first == :title || related_resource_title?(path)) && path.include?(:structuredValue)
+          return unless hash[:value] && (structured_value_title?(path) || related_resource_title?(path))
 
           # if there is a "value" key, make sure there is also a "type" key, only for title.structuredValue
           error_paths_missing_title_type << path_to_s(path) unless hash[:type]
         end
 
         def related_resource_title?(path)
-          # title is directly within relatedResource, e.g [:relatedResource, 0, :title, 0, :structuredValue, 0])
-          path.first == :relatedResource && path[2] == :title
+          # title is within relatedResource, e.g ["relatedResource", 0, "title", 0, "structuredValue", 0])
+          structured_value_path = path[4] == 'structuredValue' || (path[4] == 'parallelValue' && path[6] == 'structuredValue')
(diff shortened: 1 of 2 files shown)


lib/cocina/models/validators/description_values_validator.rb
```diff
@@ -62,22 +62,30 @@ def validate_values_for_blanks(hash, path)
         end
 
         def validate_values_for_multiples(hash, path)
-          return unless hash.count { |key, value| %i[value groupedValue structuredValue parallelValue].include?(key) && value.present? } > 1
+          return unless hash.count { |key, value| %w[value groupedValue structuredValue parallelValue].include?(key) && value.present? } > 1
```
lwrubel, line 65: keys are strings, rather than symbols.

mjgiarlo: How do we know these keys are strings?

The entrypoint to the `DescriptionValuesValidator` class is the `Validator` class, which [converts the attributes to be validated into a `HashWithIndifferentAccess` instance](https://github.com/sul-dlss/cocina-models/blob/main/lib/cocina/models/validators/validator.rb#L24-L25).

The `HashWithIndifferentAccess` class maps symbol keys to string keys in its writing interface. See https://api.rubyonrails.org/classes/ActiveSupport/HashWithIndifferentAccess.html. Thus:
> You are guaranteed that the key is returned as a string:

mjgiarlo (approved):
Yay!

mjgiarlo:
Rebased this PR on Oct. 3, 2025, and ran a report against prod using this branch. Here are the results. There are 979 druids in prod that fail this validation, with a total of 3,150 invalid cocina values. There's a pretty even distribution of the # of errors per affected druid, with a range of 1 - 7 errors per druid.

As for what the error message distribution is:
```
   2801 "Multiple value, groupedValue, structuredValue, and parallelValue in description: title1"
     99 "Multiple value, groupedValue, structuredValue, and parallelValue in description: note1, note2, note3"
     82 "Multiple value, groupedValue, structuredValue, and parallelValue in description: note1, note2"
     56 "Multiple value, groupedValue, structuredValue, and parallelValue in description: subject2"
     44 "Multiple value, groupedValue, structuredValue, and parallelValue in description: note1"
     31 "Multiple value, groupedValue, structuredValue, and parallelValue in description: relatedResource1.title1"
     14 "Multiple value, groupedValue, structuredValue, and parallelValue in description: note1, note4"
     14 "Multiple value, groupedValue, structuredValue, and parallelValue in description: event1.date1"
      4 "Multiple value, groupedValue, structuredValue, and parallelValue in description: subject4"
      2 "Multiple value, groupedValue, structuredValue, and parallelValue in description: subject1"
      2 "Multiple value, groupedValue, structuredValue, and parallelValue in description: contributor2.name1"
      1 Bare druid may not be used as a filename or a base directory
```

arcadiafalcone:
total druids with errors: 981
fixed: 716
no error found: 70
converted from MARC, need more investigation to fix: 194
other non-descriptive error: 1

arcadiafalcone:
The MARC issue preventing vf106cq3908 from being refreshed (a missing 005 field) is different from what's causing the validation error in the note1 field - I'll try fixing that error via spreadsheet so we can move forward with this while the MARC piece gets figured out.

arcadiafalcone:
OK, I did an interim fix via spreadsheet so vf106cq3908 should pass validation now.

Issue #663, opened 2024-01-12
Add coordinate validator for map data


thatbudakguy:
Similar to #662 but related to `map` data (scanned map images) instead.

We have lots of problems with scanned map data that has coordinates that are in invalid formats; see for example:
- bb847xp3575
- bq428sj1172
- by513wd0839

@justinlittman also [made a report](https://github.com/sul-dlss/purl/issues/708#issuecomment-1782904904) that shows the breadth of coordinate data formats in scanned maps. 

This causes problems when the maps are released, but fail to index, and there's no way to know that except that they don't show up in Earthworks.

justinlittman:
@thatbudakguy Can you specify what the validation would be?

ndushay:
discussion implied:  how the validators work, and what is wished for

justinlittman:
If we created this validator, all existing invalid coordinates would need to be fixed. Does that seem feasible?

kimdurante:
Since this issue affects scanned maps and not geospatial data, I have been talking with the Maps team about possible ways to batch update records with missing or invalidate coordinates. Many older maps do not contain coordinates, or use a different projection, so we could perhaps generate approximate valid coordinates using the place name.

justinlittman:
In cocina, this would be in description > subject > type = "map coordinates"?

kimdurante:
Yes

justinlittman:
The IIIF NavPlace element is created by https://github.com/iiif-prezi/osullivan/blob/main/lib/iiif/v3/presentation/nav_place.rb.

Would the validator (1) validate that every map coordinates subject can be parsed by this code and within the valid range or (2) validate that if a map coordinates subject can be parsed by this code, it is within the valid range or (3) other?

kimdurante:
I am not familiar with this code. But I would lean towards option 2, validate that if a map coordinates subject can be parsed by this code, it is within the valid range.

lwrubel:
At what point are the invalid coordinates being created?

We initially create the map coordinates subject in the `generate-descriptive` step by taking the coordinates in the ISO19139.xml's `MD_Metadata/identificationInfo/MD_DataIdentification/extent/EX_Extent/geographicElement/EX_GeographicBoundingBox` element. There is logic in the workflow step to convert these decimal coordinates into degrees/minutes/seconds. The intent there is to create a human-readable subject in the PURL. 

[...]

PR #684, opened 2024-02-28
[HOLD] Initial map coordinates validator

lwrubel:
Refs #663. 

<s>@thatbudakguy there's probably some room to improve the clarity of the code. I made a few tweaks to the NavPlace code for it to fit here. Note that the NavPlace logic can handle extra characters in the coordinates string (it has a regex to extract the coordinates), so this validator does not care about that either. 

Also probably good to add other invalid coordinates tests. </s>

Unit tests

### DIFF
--- a/lib/cocina/models/validators/map_coordinates_validator.rb
+++ b/lib/cocina/models/validators/map_coordinates_validator.rb
@@ -1,0 +1,81 @@
+# frozen_string_literal: true
+
+require 'geo/coord'
+
+module Cocina
+  module Models
+    module Validators
+      # Validates that geospatial coordinates are parseable in DMS notation
+      class MapCoordinatesValidator
+        def self.validate(clazz, attributes)
+          new(clazz, attributes).validate
+        end
+
+        def initialize(clazz, attributes)
+          @clazz = clazz
+          @attributes = attributes
+        end
+
+        # Valid if no coordinates, or all coordinates are parseable
+        def validate
+          coordinates.map(&method(:parse))
+        end
+
+        private
+
+        attr_reader :clazz, :attributes
+
+        def druid
+          @druid ||= attributes[:externalIdentifier]
+        end
+
+        # Text of all 'subject' elements of type 'map coordinates', if any
+        def coordinates
+          @coordinates ||= (attributes.dig(:description, :subject) || [])
+                           .select { |subject| subject[:type] == 'map coordinates' }
+                           .map { |subject| subject[:value] }
+        end
+
+        # Coordinates are separated by -- or /
+        # Example: W 62°51ʹ00ʺ--W 62°04ʹ00ʺ--N 17°30ʹ20ʺ--N 16°32ʹ00ʺ
+        COORD_SEPARATOR = %r{ ?--|/}
+
+        # Attempt to parse coordinates in DMS notation
+        def parse(coordinate_text)
+          coordinate_parts = coordinate_text.split(COORD_SEPARATOR)
+          case coordinate_parts.length
+          when 2  # single point
+            [parse_coord(coordinate_parts[0], coordinate_parts[1])]
+          when 4  # bounding box
+            [parse_coord(coordinate_parts[0], coordinate_parts[2]), parse_coord(coordinate_parts[1], coordinate_parts[3])]
+          else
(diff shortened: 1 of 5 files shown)


lib/cocina/models/validators/map_coordinates_validator.rb
```diff
@@ -0,0 +1,89 @@
+# frozen_string_literal: true
+
+require 'geo/coord'
+
+module Cocina
+  module Models
+    module Validators
+      # Validates that Purl matches the external identifier (druid)
```
thatbudakguy, line 8: comment probably needs to be updated

lib/cocina/models/validators/map_coordinates_validator.rb
```diff
@@ -0,0 +1,89 @@
...
+          raise ValidationError, "Invalid map coordinates for #{druid}."
+        end
+
+        private
+
+        attr_reader :clazz, :attributes
+
+        def meets_preconditions?
+          map_coordinates
+        end
```
thatbudakguy, line 34: is this a standardized thing in validators? it's kind of an ambiguously named function, so if it's not required, i think it'd be better to just inline it as `return unless map_coordinates.any?`or something in `#validate`. but if it's part of the validator API that's fine

lib/cocina/models/validators/map_coordinates_validator.rb
```diff
@@ -0,0 +1,89 @@
...
+          @map_coordinates ||= attributes.dig(:description, :subject)&.select { |subject| subject[:type] == 'map coordinates' }
+        end
+
+        # coordinate validation from:
+        # https://github.com/iiif-prezi/osullivan/blob/9816ed3dc2fd739655523620f2602f4494310674/lib/iiif/v3/presentation/nav_place.rb#L17
+        COORD_REGEX = /(?<hemisphere>[NSEW]) (?<degrees>\d+)[°⁰*] ?(?<minutes>\d+)?[ʹ']? ?(?<seconds>\d+)?[ʺ"]?/
+
+        def invalid?(coordinate_text)
+          coordinates(coordinate_text).nil? || coordinates(coordinate_text).empty?
+        end
```
thatbudakguy, line 54: it looks like we have `activesupport` as a dependency, so I think that means you can do `coordinates(coordinate_text).blank?` which is true if it's nil or empty.

lib/cocina/models/validators/map_coordinates_validator.rb
```diff
@@ -0,0 +1,89 @@
...
+          @attributes = attributes
+        end
+
+        def validate
+          return unless meets_preconditions?
+
+          # valid condition
+          return if invalid_coordinates.empty?
+
+          raise ValidationError, "Invalid map coordinates for #{druid}."
```
thatbudakguy, line 25: if this message gets shown in the Argo UI, it might be worth including the invalid coordinates themselves so you can see what the problem is

lib/cocina/models/validators/map_coordinates_validator.rb
thatbudakguy: might need to `require 'active_support/core_ext/object/blank'`

aaron-collier (approved):

lib/cocina/models/validators/map_coordinates_validator.rb
```diff
@@ -0,0 +1,77 @@
...
+            [parse_coord(coordinate_parts[0], coordinate_parts[2]), parse_coord(coordinate_parts[1], coordinate_parts[3])]
+          else
+            raise ValidationError, "Invalid map coordinates for #{druid}: #{coordinate_text}"
+          end
+        rescue ArgumentError
+          raise ValidationError, "Invalid map coordinates for #{druid}: #{coordinate_text}"
+        end
+
+        # Coordinate validation from:
+        # https://github.com/iiif-prezi/osullivan/blob/main/lib/iiif/v3/presentation/nav_place.rb
```
jmartin-sul, line 59: small and totally nitpicky suggestion: it might be nice to link to a specific commit hash here, instead of `main`, just in case the referenced repository shifts while this code stays the same?

thatbudakguy:
superseded by #688

PR #710, opened 2024-05-17
Remove depenency on RSS gem

jcoyne:

Write our own (better) w3cdtf validator. Drop an unnecessary dependency.


mjgiarlo (approved):
I can't tell how close this work gets us to what the `rss` dependency was doing for us before, but this otherwise LGTM (and it's good to get rid of that dependency).

PR #715, opened 2024-05-22
allow factory to set reading order

peetucket:

Allow specs using factories to set the reading order when building a Dro.

Useful for cases like https://github.com/sul-dlss/pre-assembly/pull/1453


peetucket:
Not necessary.  Can patch in as needed in consumer.

PR #717, opened 2024-05-30
Update documentation with MFA-related changes

mjgiarlo:
# Why was this change made?

The docs should reflect current reality.

# How was this change tested?

I ran the commands I changed.

### DIFF
--- a/README.md
+++ b/README.md
@@ -53,10 +53,10 @@ Beyond what is necessary to test the generator, the Cocina model classes are not
 
 ## Testing validation changes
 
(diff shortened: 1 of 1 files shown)


justinlittman (approved):

README.md
```diff
@@ -94,33 +94,30 @@ RAILS_ENV=production bin/validate-cocina -p 8
...
+    ```
+1. Connect to the desired database by setting the environment variables as described in the section above.  This must be done each time you SSH back into the box to run a new report.
+1. Run the report (good idea to do it in a screen or via background process in case you get disconnected):
+    ```shell
+    bundle exec bin/rails r -e production "BadIso8601Dates.report" > BadIso8601Dates.csv
+    ```
+1. When done, you can pull the report to your laptop as needed:
+    ```shell
+    scp sdr-infra:/opt/app/deploy/dor-services-app/BadIso8601Dates.csv .
+    ```
```
mjgiarlo, line 120: Re-markdowned this so we don't have to care about correct numbering ourselves. Markdown will do this for us.

PR #721, opened 2024-06-03
Remove releaseTags from cocina-models

### LINKED ISSUE #714: Remove releaseTags
Blocked by https://github.com/sul-dlss/purl/pull/1002

This should probably be a major version of cocina-models.

aaron-collier:
Fixes #714 by removing the releaseTags from openapi.yml and it's uses.

Validation and migrations have been run.


justinlittman (commented):
I've surprised this didn't get rid of`lib/cocina/models/release_tag.rb`. However, I'm not mistaken we use that class in apps (purl-fetcher?).

aaron-collier:
@justinlittman - Good catch, I honestly just missed it because it didn't come up when grepping for `releaseTag`. However, it makes me wonder that if the model is being used by purl-fetcher, wouldn't we still need this in the api spec and validation?

I think this should be held until we decide on remediation and maybe this purl-fetcher use in any case.

Thanks.

jcoyne:
Purl-fetcher does not use Cocina::Models::ReleaseTag

justinlittman:
Looks like Argo and DSA do.

aaron-collier:
Looks like it's used in hydra_etd as well. Hydra_etd already has DSC as a dependency so I should easily be able to update that. But based on our deploy repos, I think those are the only locations it's used.

justinlittman (approved):

PR #733, opened 2024-07-31
Adds reserved name validator.

### LINKED ISSUE #732: do not allow a file or folder in the root directory of a deposit to have the same name as the druid for that deposit
I'm putting this ticket in DSA because I'm not sure where to file it. We probably would do best to make sure deposits don't contain a name collision earlier in the deposit process (H2, Preassembly, ?).

----

In the AWFL layout, we use this pattern for the content-addressed storage:

`/stacks/bc/123/df/4567/bc123df4567`

The original stacks layout puts files in:

`/stacks/bc/123/df/4567`

This means there's a name collision if the depositor names a file or folder `bc123df4567` and places that in the root directory of their deposit. That would result in the depositor's own content being at the path

`/stacks/bc/123/df/4567/bc123df4567`

The results of this collision range from

- failure to deposit, if there's a druid-named file - [example](https://sul-h2-stage.stanford.edu/works/497))
- confusion over which files are content and which files are metadata, if there's a druid-named folder ([example](https://sul-h2-stage.stanford.edu/works/496))

There doesn't seem to be a problem if there's a druid-named folder or file outside of the root of the deposited content ([example](https://sul-h2-stage.stanford.edu/works/498))

I tested with user-versioned items starting in H2 but since this is at the stacks layout level, the results would be the same for hierarchical deposits coming from Preassembly.

justinlittman:
closes #732

Prevent issues with Stacks file layout.

Unit

### DIFF
--- a/lib/cocina/models/validators/reserved_filename_validator.rb
+++ b/lib/cocina/models/validators/reserved_filename_validator.rb
@@ -1,0 +1,60 @@
+# frozen_string_literal: true
+
+module Cocina
+  module Models
+    module Validators
+      # Validates that bare external identifier not used as a root filename or directory.
+      # This is necessary due to the bare druid being used as part of the Stacks file layout.
+      class ReservedFilenameValidator
+        def self.validate(clazz, attributes)
+          new(clazz, attributes).validate
+        end
+
+        def initialize(clazz, attributes)
+          @clazz = clazz
+          @attributes = attributes
+        end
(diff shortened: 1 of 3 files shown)


justinlittman:
Validation tests run in QA and stage; running in prod. I'll make sure it passes before releasing gem.

jcoyne (approved):

PR #751, opened 2024-11-20
Add field for image description

jcoyne:

This helps support accessibility and AI feature extraction for indexing

### DIFF
--- a/lib/cocina/models/presentation.rb
+++ b/lib/cocina/models/presentation.rb
@@ -4,6 +4,8 @@ module Cocina
   module Models
     # Presentation data for the File.
     class Presentation < Struct
+      # Descriptive text about an image. This can be used as alt text or for indexing. This should be 130 characters or fewer as this is best practice for alt text.
+      attribute? :description, Types::Strict::String
       # Height in pixels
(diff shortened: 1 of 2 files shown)


edsu:
Is there a reason why this isn't done a level up next to `label`?

jcoyne:
@edsu because it's only on images and it's for presentation.

andrewjbtw:
1. Won't alt-text be needed on other types of content?
2. How will someone edit this field? I think it's a good idea to add fields that users will need to edit only in conjunction with some way for a user to edit them.

jcoyne:
> How is this different than the file label field?

The label field is user provided text, but it is not intended to be used as alt text.

> Won't alt-text be needed on other types of content?

The alt attribute is only supported on images (also `<area>` and `<input>` but we don't use these to present SDR objects)

> How will someone edit this field? I think it's a good idea to add fields that users will need to edit only in conjunction with some way for a user to edit them.

Currently undetermined.

mjgiarlo:
@jcoyne @andrewjbtw Should we make time to discuss this change further or 🤷🏻‍♂️ ❓

PR #762, opened 2025-01-07
Drop Openapi3Parser dependency

mjgiarlo:
# Why was this change made?

Apparently we do not need it, and it is causing grief by way of commonmarker.

# How was this change tested?

CI

### DIFF
--- a/Gemfile.lock
+++ b/Gemfile.lock
@@ -3,7 +3,6 @@ PATH
   specs:
     cocina-models (0.99.2)
       activesupport
-      commonmarker (~> 2.0, != 2.0.2)
       deprecation
       dry-struct (~> 1.0)
       dry-types (~> 1.1)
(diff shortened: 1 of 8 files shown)


Gemfile.lock
```diff
@@ -3,7 +3,6 @@ PATH
   specs:
     cocina-models (0.99.2)
       activesupport
-      commonmarker (~> 2.0, != 2.0.2)
```
mjgiarlo, line 6: 👋🏻

cocina-models.gemspec
```diff
@@ -38,7 +33,6 @@ Gem::Specification.new do |spec|
   spec.add_dependency 'i18n' # for validating BCP 47 language tags, according to RFC 4646
   spec.add_dependency 'jsonpath' # used for date/time validation
   spec.add_dependency 'nokogiri'
-  spec.add_dependency 'openapi3_parser' # Parsing openapi doc
```
mjgiarlo, line 41: Turns out we only need one OpenAPI parser?

justinlittman (approved):

PR #789, opened 2025-04-14
Add partLabel and sortKey to catalogLinks

### LINKED ISSUE #785: Add new properties for digital serials data
Add two new properties to hold data to support serials. These new fields will be optional attributes of a `FolioCatalogLink`. Only valid for DROs/RequestDROs. 

- partLabel (string)
- sortKey  (string)

`sortKey` should only be populated if there is a `partLabel`, so I think we would need a validator to check for this. 

As background, these fields will hold data that formerly lived in description as values with types of `part number`, `part name` and `date/sequential designation`. Since they will continue to be valid and used for other purposes, don't remove `part name`,  `part number`, or `date/sequential designation` from `description_types.yml`.

Be sure to generate models from openapi.yml.

lwrubel:

Resolves #785 to add digital serials fields to catalogLinks. Models were generated. 

Unit test added for validation.

### DIFF
--- a/lib/cocina/models/validators/catalog_links_validator.rb
+++ b/lib/cocina/models/validators/catalog_links_validator.rb
@@ -21,6 +21,7 @@ module Cocina
 
           validate_catalog('symphony')
           validate_catalog('folio')
+          validate_sort_key
         end
 
         private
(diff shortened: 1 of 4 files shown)


lib/cocina/models/validators/catalog_links_validator.rb
```diff
@@ -36,6 +37,16 @@ def validate_catalog(catalog)
                                  "(only one allowed) #{refresh_catalog_links}"
         end
 
+        def validate_sort_key
```
justinlittman, line 40: JSON scheme supports `dependentRequired` that allows for this sort of validation (if you have this field, then you also must have this other field). Wondering if that can be used with OpenApi.

lwrubel: With any new keywords, we would need to update the generators as well?

lwrubel: This looks like a blocker however: `openapi_parser` does not yet support OpenAPI 3.1. https://github.com/ota42y/openapi_parser/issues/152

mjgiarlo (approved):

lib/cocina/models/validators/catalog_links_validator.rb
justinlittman: Why can't we have nice things?

PR #792, opened 2025-04-18
Add partLabel to title for display.

### LINKED ISSUE #790: Use new serials fields in title display (will support Argo display)
Update display in Argo to use `partLabel` and `sortKey` in the title display, when those fields are present. See [DMV records](https://argo.stanford.edu/view/wf604fx5181), where the structuredTitle's part name and part number info is appended to the title.

lwrubel:
Resolves #790. Once merged, will need to do https://github.com/sul-dlss/dor-services-app/issues/5311 and then test indexing for Argo display. Argo should not need changes, although we will need sample data. 

Unit. 

### DIFF
--- a/lib/cocina/models/builders/title_builder.rb
+++ b/lib/cocina/models/builders/title_builder.rb
@@ -23,7 +23,7 @@ module Cocina
                              'It must be called with an array of titles')
             titles = titles.description.title
           end
-          new(strategy: strategy, add_punctuation: add_punctuation).build(titles, part_label: part_label)
+          new(strategy: strategy, add_punctuation: add_punctuation, part_label: part_label).build(titles)
         end
 
         # the "main title" is the title withOUT subtitle, part name, etc.  We want to index it separately so
@@ -59,10 +59,12 @@ module Cocina
         #   of primary, untyped, first occurrence.  When false, return an array containing all the parallel values.
         #   Why? Think of e.g. title displayed in blacklight search results vs boosting values for ranking of search
         #   results
-        def initialize(strategy:, add_punctuation:, only_one_parallel_value: true)
+        # @param part_label [String] the partLabel to add for digital serials display
+        def initialize(strategy:, add_punctuation:, only_one_parallel_value: true, part_label: nil)
           @strategy = strategy
           @add_punctuation = add_punctuation
           @only_one_parallel_value = only_one_parallel_value
+          @part_label = part_label
         end
 
         # @param [Array<Cocina::Models::Title>] cocina_titles the titles to consider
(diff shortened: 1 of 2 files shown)


lib/cocina/models/builders/title_builder.rb
```diff
@@ -8,42 +8,44 @@ module Builders
...
-        # @param [[Array<Cocina::Models::Title,Cocina::Models::DescriptiveValue>] titles the titles to consider
+        # @param [Array<Cocina::Models::Title,Cocina::Models::DescriptiveValue>] titles the titles to consider
+        # @param [Array<Cocina::Models::FolioCatalogLink>] catalog_links the folio catalog links to check for digital serials part labels
         # @param [Symbol] strategy ":first" is the strategy for selection when primary or display
         #   title are missing
         # @param [Boolean] add_punctuation determines if the title should be formmated with punctuation
         # @return [String, Array] the title value for Solr - for :first strategy, a string; for :all strategy, an array
         #   (e.g. title displayed in blacklight search results vs boosting values for search result rankings)
-        def self.build(titles, strategy: :first, add_punctuation: true)
+        def self.build(titles, catalog_links: [], strategy: :first, add_punctuation: true)
```
mjgiarlo, line 18: I like injecting the optional catalog links instead of tossing the whole Cocina at this method.

lib/cocina/models/builders/title_builder.rb
```diff
@@ -8,42 +8,44 @@ module Builders
       # TitleBuilder selects the prefered title from the cocina object for solr indexing
       class TitleBuilder # rubocop:disable Metrics/ClassLength
         extend Deprecation
-        # @param [[Array<Cocina::Models::Title,Cocina::Models::DescriptiveValue>] titles the titles to consider
+        # @param [Array<Cocina::Models::Title,Cocina::Models::DescriptiveValue>] titles the titles to consider
+        # @param [Array<Cocina::Models::FolioCatalogLink>] catalog_links the folio catalog links to check for digital serials part labels
```
mjgiarlo, line 12: Non-blocking question: whose job should be to protect against passing Symphony catalog links along in this argument? Or is that a far-fetched or impossible scenario?

lib/cocina/models/builders/title_builder.rb
```diff
@@ -63,17 +65,18 @@ def initialize(strategy:, add_punctuation:, only_one_parallel_value: true)
...
 
-        # @param [[Array<Cocina::Models::Title>] cocina_titles the titles to consider
+        # @param [Array<Cocina::Models::Title>] cocina_titles the titles to consider
+        # @param [String, nil] part_label the partLabel to add to the title for digital serials
         # @return [String, Array] the title value for Solr - for :first strategy, a string; for :all strategy, an array
         #   (e.g. title displayed in blacklight search results vs boosting values for search result rankings)
         #
         # rubocop:disable Metrics/PerceivedComplexity
-        def build(cocina_titles)
+        def build(cocina_titles, part_label: nil)
```
mjgiarlo, line 74: We could consider injecting either the `part_label` or the `catalog_links` in the `.new` call on line 26 and making it an `attr_reader` on the instance to avoid having to pass it around and default it in a couple places, but I don't think that's a biggie.

lib/cocina/models/builders/title_builder.rb
lwrubel: Line 19 is supposed to make sure we're only dealing with Folio links. It's not very likely there would be a partLabel in Symphony--something would really have had to go wrong--but at least we're checking here. Seems to make sense here rather than all the other possible places.

lib/cocina/models/builders/title_builder.rb
lwrubel: Took this approach, thanks for the suggestion!

PR #851, opened 2025-09-29
Bump rubocop ruby target version to 3.4

aaron-collier:

A bit of style cleanup to catch up with ruby 3.4 styling.

### DIFF
--- a/cocina-models.gemspec
+++ b/cocina-models.gemspec
@@ -22,7 +22,7 @@ Gem::Specification.new do |spec|
   spec.bindir        = 'exe'
   spec.executables   = spec.files.grep(%r{^exe/}) { |f| File.basename(f) }
   spec.require_paths = ['lib']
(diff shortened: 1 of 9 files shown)


cocina-models.gemspec
```diff
@@ -22,7 +22,7 @@ Gem::Specification.new do |spec|
   spec.bindir        = 'exe'
   spec.executables   = spec.files.grep(%r{^exe/}) { |f| File.basename(f) }
   spec.require_paths = ['lib']
-  spec.required_ruby_version = '>= 3.0'
+  spec.required_ruby_version = '>= 3.4'
```
mjgiarlo, line 25: Seems safe since:

1. We keep up with Ruby versions
2. We do not use different versions of Ruby across our portfolio
3. It's extremely unlikely anyone but our team uses the gem

PR #860, opened 2025-10-22
Handle a single parallel name when mapping to MODS.

justinlittman:
refs https://github.com/sul-dlss/dor-services-app/issues/5582

Avoid indexing error

Unit


spec/cocina/models/mapping/to_mods/name_writer_spec.rb
```diff
@@ -100,4 +100,49 @@
...
+                    },
+                    valueScript: {
+                      code: 'Cyrl',
+                      source: {
+                        code: 'iso15924'
+                      }
+                    }
+                  }
+                }
+              ]
```
mjgiarlo: @justinlittman Is this a valid parallel value?

PR #863, opened 2025-10-24
Add the last batch of known DOI exceptions per Amy Hodge

mjgiarlo:
Account for all known DOIs in SDR objects so they may be edited.


mjgiarlo:
Bogus coverage ding, I think?

Issue #835, opened 2025-09-03
Validate descriptive Cocina to ensure that main value is present


arcadiafalcone:
Cocina description contains many supplementary attributes that are only relevant if the main value of the property is present. For example, providing an "encoding" is irrelevant unless there is a date value for the encoding to apply to. Users may inadvertently create "orphan" attributes without a main value in the Cocina spreadsheet, resulting in messy data that can be challenging to parse. Validation should identify these orphan attributes for removal, either automatically or by prompting the user (TBD with @andrewjbtw input).

See https://github.com/sul-dlss/cocina-models/issues/830 for some examples of orphan attributes in Cocina.

Below are listed the "main value" subproperties for Cocina properties. At least one of these subproperties should be present for the parent property to be valid.

DescriptiveValue/DescriptiveBasicValue
* value
* code
* uri
* identifier (followed by at least one valid DescriptiveValue)
* note (followed by at least one valid DescriptiveValue))
* valueAt
* structuredValue/parallelValue/groupedValue (followed by at least one valid DescriptiveValue)

Contributor
* name (followed by at least one valid DescriptiveValue)
* identifier (followed by at least one valid DescriptiveValue)
* valueAt

Event
* date (followed by at least one valid DescriptiveValue)
* contributor (followed by at least one valid Contributor)
* location (followed by at least one valid DescriptiveValue)
* identifier (followed by at least one valid DescriptiveValue)
* note (followed by at least one valid DescriptiveValue)
* structuredValue (followed by at least one valid DescriptiveValue)

Language
* value
* code
* uri
* note (followed by at least one valid DescriptiveValue)
* script (followed by at least one valid DescriptiveValue)
* valueAt
* structuredValue/parallelValue/groupedValue (followed by at least one valid DescriptiveValue)

DescriptiveAccessMetadata
* any property (followed by at least one valid DescriptiveValue)

DescriptiveGeographicMetadata
* form (followed by at least one valid DescriptiveValue)
* subject (followed by at least one valid DescriptiveValue)

RelatedResource
* title (followed by at least one valid DescriptiveValue)
* note (followed by at least one valid DescriptiveValue)
* identifier (followed by at least one valid DescriptiveValue)
* valueAt

DescriptiveAdminMetadata
* contributor (followed by at least one valid Contributor)
* event (followed by at least one valid Event)
* language (followed by at least one valid Language)
* note (followed by at least one valid DescriptiveValue)
* metadataStandard (followed by at least one valid Standard)
[...]

PR #877, opened 2025-11-18
Switch to use json_schemer

jcoyne:
It purports to support OpenAPI 3.1.

One change is that it's interpretation of how allOf + additionalFields: false works, is different than the previous approach, so I have removed "additionalFields: false" from the schema.

Part of https://github.com/sul-dlss/cocina-models/issues/835

### DIFF
--- a/Gemfile.lock
+++ b/Gemfile.lock
@@ -9,10 +9,10 @@ PATH
       edtf
       equivalent-xml
       i18n
+      json_schemer (~> 2.0)
       jsonpath
       marc (~> 1.3)
       nokogiri
-      openapi_parser (~> 1.0)
       super_diff
       thor
       zeitwerk (~> 2.1)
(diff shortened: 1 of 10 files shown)


aaron-collier:
Cool, used json_schemer in datacite-ruby and I like the validation a lot better.

lwrubel:
Running on prod did not raise any new issues with json_schemer.

jcoyne:
I'm able to find uri values that are empty strings by running:
```SQL
SELECT ro.external_identifier as object_druid
FROM repository_objects AS ro, repository_object_versions AS rov
WHERE ro.head_version_id = rov.id
AND jsonb_path_exists(rov.description, 'strict $.**.uri ? (@ == "") ')
LIMIT 10;
```

jcoyne:
It looks like openapi_parser 1.0 is completely incompatible with this branch: https://github.com/sul-dlss/dor-services-app/pull/5684

PR #913, opened 2026-01-28
Refactor to use the Marc gem

jcoyne:
This reduces the chance for errors around hash based programming.


lwrubel (approved):
Thanks, this is an improvement over hash wrangling.

PR #917, opened 2026-01-29
Reference data for MARC mapping

lwrubel:
Provides reference data for building the MARC>Cocina mapping. Once all of the mapping is completed, including specs, we will likely be able to remove this. Version control of the reference data will be helpful for tracking changes as we implement.


lwrubel:
Closing since we have a strategy to get them under version control and they're being worked on. @arcadiafalcone gave 👍 . We can use a draft PR if there needs to be conversation or tracking changes.

PR #1031, opened 2026-04-09
feat: add validation to reject filenames with newline characters [t1030]

aaron-collier:

Fixes #1030 

Adds a json schema validation for newlines in file names.


lib/cocina/models/validators/newline_filename_validator.rb
```diff
@@ -0,0 +1,53 @@
...
+          @clazz = clazz
+          @attributes = attributes
+        end
+
+        def validate
+          return unless dro?
+
+          return if filenames.none? { |filename| filename.match?(/[\n\r]/) }
+
+          raise ValidationError, 'Filename contains a newline character'
```
justinlittman: Should you report which filename?

lib/cocina/models/validators/newline_filename_validator.rb
```diff
@@ -0,0 +1,54 @@
...
+          @attributes = attributes
+        end
+
+        def validate
+          return unless dro?
+
+          offending = filenames.find { |filename| filename.match?(/[\n\r]/) }
+          return unless offending
+
+          raise ValidationError, "Filename contains a newline character: #{offending.inspect}"
```
aaron-collier: `.inspect` is a deliberate choice here to include the newline characters in the error message.

lib/cocina/models/validators/newline_filename_validator.rb
aaron-collier: Fixed. Thank you.

lib/cocina/models/validators/newline_filename_validator.rb
```diff
@@ -0,0 +1,54 @@
...
+          clazz::TYPES.intersect?(DRO::TYPES) && attributes[:externalIdentifier].present?
+        rescue NameError
+          false
+        end
+
+        def filenames
+          [].tap do |filenames|
+            next if attributes.dig(:structural, :contains).nil?
+
+            attributes[:structural][:contains].each do |fileset|
```
justinlittman: If you do `Array(attributes[:structural][:contains]).each`, then `next if` isn't necessary.

Same below as well.

schema.json
```diff
@@ -1180,7 +1180,8 @@
         },
         "filename": {
           "description": "Filename for a file. Can be same as label.",
-          "type": "string"
+          "type": "string",
```
justinlittman, line 1183: Why can't the description have a newline?

lib/cocina/models/validators/newline_filename_validator.rb
justinlittman: Ignore this.

schema.json
justinlittman: Duh. Nevermind.

justinlittman (approved):

PR #1033, opened 2026-04-13
Use JSON schema to validate unexpected properties

### LINKED ISSUE #1027: Use JSON schema to validate unexpected properties
Currently, some invalid Cocina structures are not being caught by the JSON schema:

```ruby
3.4.0 :001 > ro = RepositoryObject.find_by(external_identifier: 'druid:nd815qk1992')
...
3.4.0 :002 > ro.versions.first.to_cocina
app/models/repository_object_version.rb:60:in 'RepositoryObjectVersion#to_cocina': [Cocina::Models::DRO.new] {"rel
easeTags" => [{"to" => "Searchworks", "who" => "mjgiarlo", "what" => "self", "release" => true}], "hasAdminPolicy"...
 => "druid:qc410yz8746"} (Hash) has invalid type for :administrative violates constraints ([Cocina::Models::Administrative.new] unexpected keys [:releaseTags] in Hash input failed) (Dry::Struct::Error)
```

This validation error is being caught by `Dry::Struct`, not the JSON schema validation. This has undesirable downstream effects: code that invokes validation must expect both `Cocina::Models::ValidationError` **and** `Dry::Struct::Error` (e.g., DSA `VersionsController` and `UserVersionsController` classes).

We can't lean on `additionalProperties: false` to catch these unexpected keys in Cocina because:

> The `additionalProperties` keyword only evaluates properties defined in the **same subschema object**. If you use `allOf` to combine multiple schemas, each subschema with `additionalProperties: false` will consider any property defined in the other subschemas as "additional" and fail validation.

We should explore using [`unevaluatedProperties`](https://json-schema.org/understanding-json-schema/reference/object#unevaluatedproperties) for this purpose.

mjgiarlo:
# Why was this change made?

Fixes #1027

This commit refactors the Cocina schema to use explicit mix-in (`*Mixin`) schemas while preserving strict validation of "terminal" schemas via the `unevaluatedProperties: false` mechanism from JSON schema. "Terminal" schemas are the ones with which SDR applications typically interact, namely `DRO`, `Collection`, and `AdminPolicy` (including their `Request*` and `*WithMetadata` variants). From the README:

> Our JSON Schema uses *terminal & mixin* composition strategy so we can enforce deep validation without running into **allOf** composition pitfalls. Terminal schemas (e.g., **DRO**, **Collection**, **AdminPolicy**, their **WithMetadata** variants, and **Request** variants) are the validation boundaries and generally set the `unevaluatedProperties: false` property. Reusable mixins (e.g., **DROMixin**, **CollectionMixin**, **AdminPolicyMixin**, and **AccessMixin**) are intentionally open and are composed into terminal schemas via **allOf** or **oneOf**, so these schemas **may not** take advantage of `unevaluatedProperties` to prevent unexpected properties.

That means the open schemas (the mixins) may still contain unexpected properties and pass JSON schema validation, which will lead to `Dry::Struct:Error`s being raised. How would we get rid of those? One option is to stop caring about DRY in the schema, and inline the mixins everywhere they are used. This means we will have some amount of duplication within the schema, so future edits may be more difficult to make since we have to consider if a single edit needs to be made to multiple schemas. It's also possible that there are more sophisticated ways to use JSON Schema internals to accomplish what we need. The potential cost there is generator code breakage.

Includes:

- Add `unevaluatedProperties: false` to terminal schemas in `schema.json` to reject unexpected properties at most levels, improving strictness.
- Refactor schema composition to use mixin/terminal pattern, ensuring only terminal schemas enforce property boundaries
- Update generator and validator code to align with new schema structure and validation boundaries.
- Add and expand tests to verify rejection of unknown properties at all nesting levels for DRO, Collection, AdminPolicy, and their variants
- Update documentation to explain schema composition and validation strategy, clarifying mixin openness and terminal strictness
- Regenerate all models and docs

# How was this change tested?

- [x] CI
- [x] stage

### DIFF
--- a/lib/cocina/generator/generator.rb
+++ b/lib/cocina/generator/generator.rb
@@ -102,8 +102,10 @@ module Cocina
       end
 
       def schema_for(schema_name, lite: false)
+        return if schema_name.ends_with?('Mixin')
+
         schema_doc = schemas[schema_name]
-        return nil if schema_doc.nil?
+        return if schema_doc.nil?
 
         case schema_doc.type
         when 'object'
--- a/lib/cocina/generator/schema.rb
+++ b/lib/cocina/generator/schema.rb
@@ -5,21 +5,25 @@ module Cocina
     # Class for generating from a JSON schema
     class Schema < SchemaBase
       def schema_properties
-        @schema_properties ||= (properties + all_of_properties + one_of_properties).uniq(&:key)
+        @schema_properties ||= (
+          schema_properties_for(schema_doc) +
+          all_of_properties_for(schema_doc) +
+          one_of_properties_for(schema_doc)
+        ).uniq(&:key)
       end
 
-      VALIDATE_TYPES = %w[DRO RequestDRO DROWithMetadata Collection RequestCollection CollectionWithMetadata AdminPolicy
-                          RequestAdminPolicy AdminPolicyWithMetadata Description RequestDescription].freeze
+      VALIDATABLE_TYPES = %w[DRO RequestDRO DROWithMetadata Collection RequestCollection CollectionWithMetadata AdminPolicy
+                             RequestAdminPolicy AdminPolicyWithMetadata Description RequestDescription].freeze
 
       def generate
         <<~RUBY
           # frozen_string_literal: true
 
           module Cocina
-            module Models#{'              '}
+            module Models
               #{preamble}class #{name} < Struct
 
-                #{validate}
+                #{validatable}
                 #{types}
 
                 #{model_attributes}
@@ -65,68 +69,41 @@ module Cocina
         RUBY
       end
 
-      def validate
-        return '' unless validatable?
+      def validatable
+        return '' if VALIDATABLE_TYPES.exclude?(name)
 
         <<~RUBY
           include Validatable
         RUBY
       end
 
-      def validatable?
-        VALIDATE_TYPES.include?(name) && !lite
-      end
-
-      def properties
-        schema_properties_for(schema_doc)
-      end
-
-      def all_of_properties
-        all_of_properties_for(schema_doc)
-      end
-
-      def one_of_properties
-        one_of_properties_for(schema_doc)
-      end
-
       def all_of_properties_for(doc)
-        return [] if doc.all_of.nil?
-
-        doc.all_of.map do |all_of_schema|
-          # All of for this + recurse
+        Array(doc.all_of).flat_map do |all_of_schema|
           schema_properties_for(all_of_schema) +
             all_of_properties_for(all_of_schema) +
             one_of_properties_for(all_of_schema)
-        end.flatten
-      end
-
-      def one_of_properties_for(doc)
-        return [] if doc.one_of.nil?
-
-        # All properties must be objects.
-        raise 'All properties for oneOf must be objects' unless doc.one_of.all? { |schema| schema.type == 'object' }
-
-        doc.one_of.flat_map do |one_of_doc|
-          one_of_doc.properties.map do |key, properties_doc|
-            property_class_for(properties_doc).new(properties_doc,
-                                                   key: key,
-                                                   # The property does less validation because may vary between
-                                                   # different oneOf schemas. Validation is still performed
-                                                   # by JSON Schema.
-                                                   relaxed: true,
-                                                   parent: self,
-                                                   schemas: schemas)
-          end
         end
       end
 
-      def schema_properties_for(doc)
+      def one_of_properties_for(doc)
+        Array(doc.one_of).flat_map do |one_of_schema|
+          schema_properties_for(one_of_schema, relaxed: true) +
+            all_of_properties_for(one_of_schema) +
+            one_of_properties_for(one_of_schema)
+        end
+      end
+
+      def schema_properties_for(doc, relaxed: nil)
+        relax_all_properties = relaxed
+
         Array(doc.properties).map do |key, properties_doc|
           clazz = property_class_for(properties_doc)
+          relaxed = relax_all_properties.nil? ? lite && clazz != SchemaValue : relax_all_properties
+
           clazz.new(properties_doc,
                     key: key,
                     required: doc.required&.include?(key),
-                    relaxed: lite && clazz != SchemaValue,
+                    relaxed: relaxed,
                     nullable: Array(properties_doc.type).include?('null'),
                     parent: self,
                     schemas: schemas)
--- a/lib/cocina/models/descriptive_value.rb
+++ b/lib/cocina/models/descriptive_value.rb
@@ -4,6 +4,7 @@ module Cocina
   module Models
     # Default value model for descriptive elements.
     class DescriptiveValue < Struct
+      attribute :appliesTo, Types::Strict::Array.of(DescriptiveBasicValue).default([].freeze)
       attribute :structuredValue, Types::Strict::Array.of(DescriptiveValue).default([].freeze)
       attribute :parallelValue, Types::Strict::Array.of(DescriptiveValue).default([].freeze)
       attribute :groupedValue, Types::Strict::Array.of(DescriptiveValue).default([].freeze)
@@ -40,7 +41,6 @@ module Cocina
       attribute? :valueLanguage, DescriptiveValueLanguage.optional
       # URL or other pointer to the location of the value of the descriptive element.
       attribute? :valueAt, Types::Strict::String
-      attribute :appliesTo, Types::Strict::Array.of(DescriptiveBasicValue).default([].freeze)
     end
   end
 end
--- a/lib/cocina/models/mapping/normalizers/mods_normalizer.rb
+++ b/lib/cocina/models/mapping/normalizers/mods_normalizer.rb
@@ -394,13 +394,14 @@ module Cocina
           end
 
           def blank_ng_xml
-            Nokogiri::XML(<<~XML
-              <mods xmlns="http://www.loc.gov/mods/v3"#{' '}
-                xmlns:xsi="http://www.w3.org/2001/XMLSchema-instance"#{' '}
-                version="3.6"#{' '}
-                xsi:schemaLocation="http://www.loc.gov/mods/v3 http://www.loc.gov/standards/mods/v3/mods-3-6.xsd" />
-            XML
-                         )
+            Nokogiri::XML(
+              <<~XML
+                <mods xmlns="http://www.loc.gov/mods/v3"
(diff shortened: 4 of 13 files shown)


lib/cocina/generator/generator.rb
```diff
@@ -102,8 +102,10 @@ def schemas
       end
 
       def schema_for(schema_name, lite: false)
+        return if schema_name.ends_with?('Mixin')
```
mjgiarlo, line 105: See the PR description and the README for more information about what `Mixin` means here. In short, it's a schema def in `schema.json` which is used only to compose other schema defs in JSONland; we do not want the corresponding Ruby class.

lib/cocina/generator/schema.rb
```diff
@@ -5,21 +5,25 @@ module Generator
...
+        @schema_properties ||= (
+          schema_properties_for(schema_doc) +
+          all_of_properties_for(schema_doc) +
+          one_of_properties_for(schema_doc)
+        ).uniq(&:key)
       end
 
-      VALIDATE_TYPES = %w[DRO RequestDRO DROWithMetadata Collection RequestCollection CollectionWithMetadata AdminPolicy
-                          RequestAdminPolicy AdminPolicyWithMetadata Description RequestDescription].freeze
+      VALIDATABLE_TYPES = %w[DRO RequestDRO DROWithMetadata Collection RequestCollection CollectionWithMetadata AdminPolicy
```
mjgiarlo, line 15: Language tweak.

lib/cocina/generator/schema.rb
```diff
@@ -5,21 +5,25 @@ module Generator
     # Class for generating from a JSON schema
     class Schema < SchemaBase
       def schema_properties
-        @schema_properties ||= (properties + all_of_properties + one_of_properties).uniq(&:key)
+        @schema_properties ||= (
+          schema_properties_for(schema_doc) +
```
mjgiarlo, line 9: The methods that were replaced were only used once, so I inlined the calls for increased clarity.

lib/cocina/generator/schema.rb
```diff
@@ -5,21 +5,25 @@ module Generator
...
+      VALIDATABLE_TYPES = %w[DRO RequestDRO DROWithMetadata Collection RequestCollection CollectionWithMetadata AdminPolicy
+                             RequestAdminPolicy AdminPolicyWithMetadata Description RequestDescription].freeze
 
       def generate
         <<~RUBY
           # frozen_string_literal: true
 
           module Cocina
-            module Models#{'              '}
+            module Models
```
mjgiarlo, line 23: I suspect this was cruft from a Rubocop-related commit.

lib/cocina/generator/schema.rb
```diff
@@ -65,68 +69,41 @@ def types
...
-        return [] if doc.all_of.nil?
-
-        doc.all_of.map do |all_of_schema|
-          # All of for this + recurse
+        Array(doc.all_of).flat_map do |all_of_schema|
           schema_properties_for(all_of_schema) +
             all_of_properties_for(all_of_schema) +
             one_of_properties_for(all_of_schema)
-        end.flatten
+        end
```
mjgiarlo, line 85: The changes to this method are a pure refactor, making it less verbose and better aligned with `#one_of_properties_for`.

lib/cocina/generator/schema.rb
```diff
@@ -65,68 +69,41 @@ def types
...
-                                                   # by JSON Schema.
-                                                   relaxed: true,
-                                                   parent: self,
-                                                   schemas: schemas)
-          end
+        Array(doc.one_of).flat_map do |one_of_schema|
+          schema_properties_for(one_of_schema, relaxed: true) +
+            all_of_properties_for(one_of_schema) +
+            one_of_properties_for(one_of_schema)
         end
```
mjgiarlo, line 93: I suspected we could simplify this method and make it look more like `#all_of_properties_for`, and that does appear to be the case once a seam is provided in `#schema_properties_for` that allows us to ensure a relaxed generation here.

lib/cocina/models/mapping/normalizers/mods_normalizer.rb
```diff
@@ -394,13 +394,14 @@ def normalize_related_item_attributes
           end
 
           def blank_ng_xml
-            Nokogiri::XML(<<~XML
-              <mods xmlns="http://www.loc.gov/mods/v3"#{' '}
-                xmlns:xsi="http://www.w3.org/2001/XMLSchema-instance"#{' '}
-                version="3.6"#{' '}
-                xsi:schemaLocation="http://www.loc.gov/mods/v3 http://www.loc.gov/standards/mods/v3/mods-3-6.xsd" />
-            XML
-                         )
```
mjgiarlo, line 403: Cleaning up what looks like a crufty Rubocop autocorrection.

lib/cocina/models/validators/json_schema_validator.rb
```diff
@@ -6,14 +6,12 @@ module Validators
       # Perform validation against JSON schema
       class JsonSchemaValidator
         def self.validate(clazz, attributes)
-          return unless clazz.name
```
mjgiarlo, line 9: Ran whole spec suite and re-generated and this is never triggered so I suspect this is dead code. Remove and simplify.

lib/cocina/models/validators/json_schema_validator.rb
```diff
@@ -6,14 +6,12 @@ module Validators
...
-
           method_name = clazz.name.split('::').last
 
-          attributes['cocinaVersion'] = Cocina::Models::VERSION if %w[DRO RequestDRO AdminPolicy RequestAdminPolicy Collection RequestCollection DROWithMetadata].include? method_name
+          attributes['cocinaVersion'] = Cocina::Models::VERSION if clazz.attribute_names.include?(:cocinaVersion)
```
mjgiarlo, line 11: Instead of maintaining a list (and an incomplete one at that) of models that support cocina version, ask the schema what its attributes are and determine based on actual behavior.

lib/cocina/models/validators/json_schema_validator.rb
```diff
@@ -6,14 +6,12 @@ module Validators
...
 
           errors = schema.ref("#/$defs/#{method_name}").validate(attributes.as_json).to_a
-          return unless errors.any?
+          return if errors.empty?
```
mjgiarlo, line 14: Prefer `if` to `unless` ... unless that makes things less clear.

lib/cocina/models/descriptive_value.rb
```diff
@@ -4,6 +4,7 @@ module Cocina
   module Models
     # Default value model for descriptive elements.
     class DescriptiveValue < Struct
+      attribute :appliesTo, Types::Strict::Array.of(DescriptiveBasicValue).default([].freeze)
```
mjgiarlo, line 7: Because of the new schema composition changes, some models have attrs in a different order. I don't *think* this is problematic but I'm open to discussion.

lib/cocina/generator/schema.rb
```diff
@@ -65,68 +69,41 @@ def types
...
+      def validatable
+        return '' if VALIDATABLE_TYPES.exclude?(name)
 
         <<~RUBY
           include Validatable
         RUBY
       end
 
-      def validatable?
-        VALIDATE_TYPES.include?(name) && !lite
```
justinlittman, line 77: Can `!lite` be dropped here?

lib/cocina/models/descriptive_value.rb
justinlittman: Wondering if this will change the serialization of the cocina which _might_ be a problem?

lib/cocina/generator/schema.rb
mjgiarlo: I thought so. If `lite` is `true`, then `name` would end in `Lite` and none of the members of `VALIDATABLE_TYPES` end in `Lite`. Does that sound right to you too?

lib/cocina/models/descriptive_value.rb
mjgiarlo: > Wondering if this will change the serialization of the cocina which _might_ be a problem?

Maybe? Do you have ideas around how I might test this concern?

justinlittman: That could be in any one of our systems. Best first cut would be to try running the specs for each of the repos pointing at this branch.

mjgiarlo: @justinlittman I created PRs in argo, common-accessioning, dor-services-app, gis-robot-suite, google-books, heracles-etd, hungry-hungry-hippo, pre-assembly, sdr-api, was-registrar-app, & was_robot_suite, and all the tests passed except for a few in DSA that flagged actual invalid cocina data lurking in tests. (I'll make a real follow-up PR in DSA if we want to keep moving this forward.)

Any other testing you'd recommend?

justinlittman: I'd look hard at roundtripping in H3 and other parts of the code where we check if a cocina object has changed. Do you know which fields are in different order?

justinlittman: I'd be surprised if this surfaced in unit OR integration tests because those are all creating new cocina objects, whereas a problem might arise with the use of existing cocina objects.

mjgiarlo: @justinlittman 
> ... Do you know which fields are in different order?

Yeah:

* `appliesTo` shifts from the end to the beginning in `DescriptiveValue`
* `appliesTo` shifts from the end to the beginning in `Title`
* `view`, `download`, `location`, and `controlledDigitalLending` shift from the beginning to the end in `DroAccess`
* `releaseDate` and `useAndReproductionStatement` shift from the end to the beginning in `Embargo`

mjgiarlo: @justinlittman 💬 
> I'd be surprised if this surfaced in unit OR integration tests because those are all creating new cocina objects, whereas a problem might arise with the use of existing cocina objects.

OK, would a useful next test be editing an object in H3 that pre-existed this work?

justinlittman: Yes, that would be a good place to start.

mjgiarlo: @justinlittman I just edited https://sul-h3-qa.stanford.edu/works/druid:cx786ds8831 with the testing branch. Any other tests?

justinlittman: Not that I can think of.

justinlittman (approved):

Issue #1036, opened 2026-04-28
Add new "405-" barcode prefix to valid barcode patterns


andrewjbtw:
The acquisition of California Historical Society materials introduced a new barcode into FOLIO and SUL's collections. See https://searchworks.stanford.edu/view/in00000326278 for an example.

CHS barcodes begin with the prefix "405". Users can't associated items with these barcodes until we add this as a valid pattern. It looks like these patterns are defined in the [Cocina schema.json](https://github.com/sul-dlss/cocina-models/blob/824ed5efa20bbf906bfd1d21b9fc5a336f122462/schema.json#L223).

jmartin-sul:
change released in https://github.com/sul-dlss/cocina-models/releases/tag/v0.117.0

PR #1037, opened 2026-04-29
Add CHS barcode format

jmartin-sul:

* add barcode format for California Historical Society (CHS)
* opportunistic redundancy reduction (stumbled on "DRO object" in other barcode descriptions, but "DRO" = "Digital Repository Object")
  * left as a second commit for easier removal in case reviewer disagrees with this change

connects with #1036 (will need to cut a release to close the ticket)

unit tests (expands space of valid barcodes, so running validation on existing cocina seemed unnecessary, but happy to test validation if others do think it's warranted)

### DIFF
--- a/lib/cocina/models/barcode.rb
+++ b/lib/cocina/models/barcode.rb
@@ -3,6 +3,6 @@
 module Cocina
   module Models
     # A barcode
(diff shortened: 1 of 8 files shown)


jmartin-sul:
converting to draft, since andrew inquired about whether the barcode length is a fixed number of digits (it is, updating the PR accordingly)

peetucket (approved):

PR #1041, opened 2026-05-11
add useful context to cocina-models partial release instructions

jmartin-sul:

so that our future selves aren't tempted to make more work by trying to skip seemingly unnecessary work (that turned out to be necessary afterall).

see https://github.com/sul-dlss/cocina-models/pull/1038/changes#r3219032947 and https://stanfordlib.slack.com/archives/C09M7P91R/p1777911928123009?thread_ts=1777672353.975549&cid=C09M7P91R

this was what worked when https://github.com/sul-dlss/cocina-models/pull/1038 didn't

### DIFF
--- a/README.md
+++ b/README.md
@@ -121,6 +121,7 @@ Before you release a major or minor change, think about if this release will inc
 If unsure, ask the team or ask for help to just run the validation report anyway (as described above).
 
 ### Partial release process
+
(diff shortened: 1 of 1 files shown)


README.md
```diff
@@ -121,6 +121,7 @@ Before you release a major or minor change, think about if this release will inc
 If unsure, ask the team or ask for help to just run the validation report anyway (as described above).
 
 ### Partial release process
+
```
jmartin-sul, line 124: whoops, this was the start of an edit i backed out of, will remove this change after lunch

PR #1043, opened 2026-05-19
Filter out unhelpful unevaluatedProperties errors

mjgiarlo:
# Why was this change made?

Currently, if there is (e.g.) one nested bogus property, it is correctly included in the error output but due to https://github.com/davishmcclurg/json_schemer/issues/157 the errors *also* include every root property as bogus even though they are not. We filter these out in cocina-models now to help everyone make sense of cocina validation errors.

# How was this change tested?

- [x] CI
- [x] Stage


justinlittman (approved):
Can you add some comments as to why filtering is necessary?

PR #1058, opened 2026-06-02
Validate main value present for RelatedResource

### LINKED ISSUE #1004: Validate main value present for RelatedResource
This ticket is part in the larger "main value" cocina-models validation work in https://github.com/sul-dlss/cocina-models/issues/835.

At least one of these sub-properties should be present for the RelatedResource property to be valid.

title (followed by at least one valid DescriptiveValue)
note (followed by at least one valid DescriptiveValue)
identifier (followed by at least one valid DescriptiveValue)
valueAt

After implementing the changes to the schema to support this validation, run bin/validate-cocina to provide @andrewjbtw and @arcadiafalcone with information about any needed remediation. The work should remain on a branch until remediation is done.

mjgiarlo:
# Why was this change made?

Fixes #1004

# How was this change tested?

CI, ran validation multiple times. Per Arcadia, the only violations of the new validations in prod should be objects that will be fixed by refreshing from ILS records.

### DIFF
--- a/schema.json
+++ b/schema.json
@@ -2150,7 +2150,7 @@
           "description": "Titles of the related resource.",
           "type": "array",
           "items": {
-            "$ref": "#/$defs/DescriptiveValue"
+            "$ref": "#/$defs/Title"
           }
         },
         "contributor": {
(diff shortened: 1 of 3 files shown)


schema.json
```diff
@@ -2150,7 +2150,7 @@
           "description": "Titles of the related resource.",
           "type": "array",
           "items": {
-            "$ref": "#/$defs/DescriptiveValue"
+            "$ref": "#/$defs/Title"
```
mjgiarlo, line 2153: Be more specific so that titles are validated as titles.

justinlittman (approved):

PR #1062, opened 2026-06-05
Extract description validators to use visitor pattern.

justinlittman:
refs #1051

Faster performance, especially getting rid of JSONPath.

This was tested with `bin/validate-data` using a recent dump.


lib/cocina/models/mapping/from_mods/event.rb
```diff
@@ -435,7 +435,7 @@ def build_structured_date(date_nodes)
               new_node = node.deep_dup
               new_node.remove_attribute('encoding') if common_attribs[:encoding].present? || node[:encoding]&.empty?
               new_node.remove_attribute('qualifier') if common_attribs[:qualifier].present? || node[:qualifier]&.empty?
-              build_date(new_node)
+              build_date(new_node, encoding: common_attribs.dig(:encoding, :code))
```
justinlittman, line 438: This fixes a bug in the code for handling of EDTF dates that have less than 4 digits.

lib/cocina/models/validators/validator.rb
```diff
@@ -7,23 +7,21 @@ module Validators
...
           AssociatedNameValidator,
-          DescriptionTypesValidator,
-          DescriptionValuesValidator,
-          DateTimeValidator,
           LanguageTagValidator,
           ReservedFilenameValidator
         ].freeze
 
-        def self.validate(clazz, attributes)
+        def self.validate(clazz, attributes, validators: VALIDATORS)
```
justinlittman, line 17: This will make it easier to do testing in the console in the future.

lib/cocina/models/mapping/from_mods/event.rb
```diff
@@ -489,6 +491,12 @@ def clean_date(date)
             date.delete_suffix('.')
           end
 
+          def pad_edtf_year(value)
```
mjgiarlo, line 495: Can you add a comment about the intent behind the regex here?

justinlittman: Definitely.

mjgiarlo (approved):

PR #1087, opened 2026-06-15
Adds resource types validator.

justinlittman:
closes #1085

Per @arcadiafalcone 

`bin/validate-data cocina-head-versions-prod-2026-06-15.jsonl.xz`

### DIFF
--- a/lib/cocina/models/validators/description_form_resource_type_visitor_validator.rb
+++ b/lib/cocina/models/validators/description_form_resource_type_visitor_validator.rb
@@ -1,0 +1,46 @@
+# frozen_string_literal: true
+
+module Cocina
+  module Models
+    module Validators
+      # Validates form.value against allowed values when form.source.value is a known controlled vocabulary.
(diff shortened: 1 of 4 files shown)


lib/cocina/models/validators/description_form_resource_type_visitor_validator.rb
```diff
@@ -0,0 +1,46 @@
+# frozen_string_literal: true
+
+module Cocina
+  module Models
+    module Validators
+      # Validates form.value against allowed values when form.source.value is a known controlled vocabulary.
+      class DescriptionFormResourceTypeVisitorValidator < BaseDescriptionVisitorValidator
```
peetucket, line 7: curious as to the nomenclature of "visitor" ... does this have any significance and or broader meaning?

justinlittman: https://en.wikipedia.org/wiki/Visitor_pattern

peetucket (approved):

PR #1100, opened 2026-06-15
Add validator for location source code

peetucket:

Fixes #1094 - add validation for location source codes

Specs


spec/fixtures/description/etd.json
```diff
@@ -256,7 +256,7 @@
           "value": "Stanford (Calif.)",
           "uri": "http://id.loc.gov/authorities/names/n50046557",
           "source": {
-            "code": "lcnaf",
+            "code": "naf",
```
peetucket, line 259: turns out an unrelated fixture had an invalid location source code which caused a test to fail with this new validation... after checking with Arcadia, we decided to update the fixture to an allowed value

justinlittman (approved):

PR #1101, opened 2026-06-15
Add validator for contributor role source code

peetucket:

Fixes #1099 - add validation for contributor role source code

~~HOLD for remediation needed (see comment in #1099)~~

Specs

### DIFF
--- a/lib/cocina/models/validators/description_role_source_code_visitor_validator.rb
+++ b/lib/cocina/models/validators/description_role_source_code_visitor_validator.rb
@@ -34,11 +34,10 @@ module Cocina
             path.any? { |part| part.to_s == 'contributor' } # there is a contributor in the path (any? allows for nested roles)
         end
 
-        # rubocop:disable Style/ClassVars
+        # Source codes allowed for contributor.role.source.code
         def valid_codes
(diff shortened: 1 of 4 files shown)


lib/cocina/models/validators/description_role_source_code_visitor_validator.rb
```diff
@@ -0,0 +1,45 @@
...
+        def role_path?(path)
+          path.length >= 2 && # ensure we have at least two elements (at minimum role and its index)
+            path[-1].is_a?(Integer) &&
+            path[-2].to_s == 'role' && # we have a nested role in the path
+            path.any? { |part| part.to_s == 'contributor' } # there is a contributor in the path (any? allows for nested roles)
+        end
+
+        # rubocop:disable Style/ClassVars
+        def valid_codes
+          @@valid_codes ||= YAML.load_file(::File.expand_path('../../../../role_source_codes.yml', __dir__)).to_set(&:downcase)
```
justinlittman: Given that there are only 4 values, would recommend putting directly in this class instead of yml file.

peetucket: had the same thought initially, this would be more consistent with other classes, happy either way though

peetucket: moved into class

justinlittman (approved):

PR #1111, opened 2026-06-16
Validate language script source codes in schema

peetucket:

Fixes #1108 - validate language script codes

~~HOLD to verify invalid objects~~

Note: gemini AI assisted code

Specs
Validation run again and turned up same objects (which have since been remediated)


lib/cocina/models/validators/description_script_source_code_visitor_validator.rb
```diff
@@ -0,0 +1,63 @@
+# frozen_string_literal: true
+
+module Cocina
+  module Models
+    module Validators
+      # Validates that language.script.source.code and valueLanguage.valueScript.source.code
```
justinlittman: Can this be enforced in the schema?

peetucket: revising and testing

schema.json
```diff
@@ -1423,7 +1423,25 @@
...
+            },
+            {
+              "type": "object",
+              "properties": {
+                "source": {
+                  "type": "object",
+                  "properties": {
+                    "code": {
+                      "type": "string",
+                      "pattern": "^[iI][sS][oO]15924$"
```
peetucket, line 1438: this pattern allows for case insensitive matches

lib/cocina/models/validators/description_script_source_code_visitor_validator.rb
peetucket: Changed to schema validation.  Ran full validation again and it turned up same objects (which have since been remediated).

Could wait to run against tomorrows dump to ensure validation passes if we want.

justinlittman (approved):

schema.json
lwrubel: Does this pattern allow values with mixed case like `iSo15294` or `isO15294` and is that OK?

peetucket: Ahh, yes, it would.  @arcadiafalcone while it would be weird to have something like "iSo15294", that would be valid with this code.   She we be case insensitive but case consistent?  Or does it not really matter?

arcadiafalcone: We can let it not matter.

PR #1118, opened 2026-06-17
Validate main value present for DescriptiveAdminMetadata.

justinlittman:
closes #1114

Per ticket

`bin/validate-data cocina-head-versions-prod-2026-06-17.jsonl.xz`


mjgiarlo (approved):
Worth adding some tests?

PR #1132, opened 2026-06-25
add validation for language and valuelanguage codes

peetucket:

Fixes #1123 - add validator for language and valuelanguage codes

Note: Claude assisted code.

Specs

### DIFF
--- a/spec/cocina/models/description_spec.rb
+++ b/spec/cocina/models/description_spec.rb
@@ -53,7 +53,7 @@ RSpec.describe Cocina::Models::Description do
       expect(language.value).to eq('English')
       expect(language.code).to eq('eng')
       expect(language.uri).to eq('http://id.loc.gov/vocabulary/iso639-2/eng')
-      expect(language.source.code).to eq('iso239-2b')
+      expect(language.source.code).to eq('iso639-2b')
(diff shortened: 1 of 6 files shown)


spec/fixtures/description/etd.json
```diff
@@ -574,7 +574,7 @@
         "code": "eng",
         "uri": "http://id.loc.gov/vocabulary/iso639-2/eng",
         "source": {
-          "code": "iso239-2b",
+          "code": "iso639-2b",
```
peetucket, line 577: these were changed because the fixture had invalid codes, which was triggering the new validator

spec/cocina/models/description_spec.rb
```diff
@@ -53,7 +53,7 @@
       expect(language.value).to eq('English')
       expect(language.code).to eq('eng')
       expect(language.uri).to eq('http://id.loc.gov/vocabulary/iso639-2/eng')
-      expect(language.source.code).to eq('iso239-2b')
+      expect(language.source.code).to eq('iso639-2b')
```
peetucket, line 56: had to update the fixture to give a valid code (see notes below), and thus the test that uses the fixture was updated

justinlittman (approved):

PR #1151, opened 2026-07-10
Improve readability of error messages from JSON schema validation

lwrubel:
Refs sul-dlss/argo#1145 and (partially) https://github.com/sul-dlss/cocina-models/issues/1164.

Reformats validation error messages produced by `jsonschema_rs` into incrementally more readable and less redundant text. Continues in the same vein as existing de-noising functions. I used Claude to work through how to parse the jsonschema error message text, but added comments to help make what is happening more transparent. 

This would change:
```[] has less than 1 item at /event/0/date, "contributor" is a required property at /event/0, "location" is a required property at /event/0, "identifier" is a required property at /event/0, "note" is a required property at /event/0, "structuredValue" is a required property at /event/0, "parallelEvent" is a required property at /event/0 ```
to:
```/event/0/date is empty but should have at least 1 item; /event/0 needs to include at least one of the following: contributor, location, identifier, note, structuredValue, parallelEvent```

CI.

### DIFF
--- a/lib/cocina/models/validators/json_schema_validator.rb
+++ b/lib/cocina/models/validators/json_schema_validator.rb
@@ -67,7 +67,7 @@ module Cocina
           evaluation = self.class.validator_for(method_name).evaluate(attributes.as_json)
           return if evaluation.valid?
 
-          raise ValidationError, "When validating #{method_name}: " + filtered_error_messages(evaluation).join(', ')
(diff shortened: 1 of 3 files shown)


mjgiarlo (approved):
Better than it was before!

justinlittman (approved):

peetucket (approved):

PR #1153, opened 2026-07-10
require title or refreshable catalog link

peetucket:

Fixes #1147 - use schema.json to require either a title or a refreshable catalog link.  Also remove some label references in the specs (not indicated in ticket, but seemed like it is the right thing to do).

Note: codex assisted code 

Specs

### DIFF
--- a/lib/cocina/models/request_admin_policy.rb
+++ b/lib/cocina/models/request_admin_policy.rb
@@ -20,7 +20,7 @@ module Cocina
       attribute(:administrative, AdminPolicyAdministrative.default { AdminPolicyAdministrative.new })
       # Description that is included in a request to create a DRO. This is the same as a
       # Description, except excludes PURL.
-      attribute? :description, RequestDescription.optional
+      attribute(:description, RequestDescription.default { RequestDescription.new })
     end
   end
 end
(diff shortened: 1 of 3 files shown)


lib/cocina/models/request_admin_policy.rb
```diff
@@ -20,7 +20,7 @@ class RequestAdminPolicy < Struct
       attribute(:administrative, AdminPolicyAdministrative.default { AdminPolicyAdministrative.new })
       # Description that is included in a request to create a DRO. This is the same as a
       # Description, except excludes PURL.
-      attribute? :description, RequestDescription.optional
+      attribute(:description, RequestDescription.default { RequestDescription.new })
```
peetucket, line 23: auto generated from schema.json

schema.json
```diff
@@ -2682,6 +2735,44 @@
...
+          "properties": {
+            "description": {
+              "required": [
+                "title"
+              ]
+            }
+          }
+        },
+        {
+          "properties": {
```
lwrubel, line 2755: The `RequestDRO` schema is slightly different from how `identification` is required for a `RequestCollection`. It needs the following, correct?
```
          "required": [
            "identification"
          ],
```

peetucket: added

lwrubel (approved):
Approved.

PR #1165, opened 2026-07-28
ensure we do not check in cocina dump file

peetucket:

Add new file format for cocina dump to gitignore

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -10,6 +10,7 @@
 # rspec failure tracking
 .rspec_status
 
+*.jsonl.zst
 *.jq.txt
 *.jsonl.xz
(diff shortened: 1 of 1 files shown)


.gitignore
```diff
@@ -10,7 +10,8 @@
 # rspec failure tracking
 .rspec_status
 
+*.jsonl.zst
 *.jq.txt
 *.jsonl.xz
 *.csv
-*-playground.html
\ No newline at end of file
+*-playground.html
```
peetucket, line 17: this is a text editor auto formatting change when i saved the file

lwrubel (approved):

PR #1178, opened 2026-07-31
update readme on how validation check occurs

peetucket:

Part of #1177 

Running validation checks will not catch all errors due to short circuiting of exceptions.  This better explains what happens now as we continue to investigate other options.


justinlittman (commented):
Can we just replace this with one sentence? "To guarantee that all errors have been remediated, generate new dumps and validate until no errors are reported."

justinlittman (approved):

PR #1179, opened 2026-07-31
collect all validation errors at once

peetucket:

See #1177 

This PR merges into the [README PR,](https://github.com/sul-dlss/cocina-models/pull/1178) since the tweaks some of the changes made to the README there.

Claude Code analyzed and suggested code (with some minor manual edits by me, and requested changes via prompts).

The `bin/validate-data` script currently stops at the first validation error that may be present in an object (after ensuring the JSON is valid) because of the way cocina-models currently handles validations.  This change is designed to pass through objects which fail validation into a separate helper method which collects the various invalid choices so they can be collected together in one pass (at the cost of efficiency).  Since this is only done for objects that are invalid, which is hopefully a relatively small number, the performance penalty should be acceptable.

New spec

### DIFF
--- a/README.md
+++ b/README.md
@@ -66,12 +66,16 @@ If there is a possibility that a model, mapping, or validation change will confl
 
 ### How validation errors are reported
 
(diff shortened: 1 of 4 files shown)


justinlittman (commented):
This is a huge amount of extra complexity / code for a corner case (an object with more than one validation problem).

